PR #56, opened 2023-10-18
feat: Extract subgraphs from supergraphs

sachindshinde:
Code for extracting subgraphs from supergraphs, and any other code in the `federation` repo that this code uses.

A few things to note:
- Despite naming, the function does not generally extract subgraphs from supergraphs, as composition at this time is not a fully reversible transformation. It extracts subgraph information for the purposes of query planning; other purposes may find this "extraction" lacking/insufficient.
- This code is not directly called by router; it's one of the early parts of query planning.
- Primarily looking for review on Rust best practices. If you're aware of the business logic in the `federation` repo and wish to review the fidelity of the ported logic, you can review that as well. (I'll leave comments on the code to point to relevant `federation` repo code.)
- This code aims to mimic the JS code where possible to ease understanding and maintenance burden, but if this leads to unacceptable Rust practices please let me know.
- There's a fair amount of boilerplate/copy-paste. I'm guessing the recommended way of avoiding this is through traits (or macros), but I figure I'd wait for recommendations before jumping to solutions.
- There's questions in places about whether to panic versus whether to return an error. My general understanding is that panics should be reserved for developer mistakes (e.g. invariant violations in data structures), while we should use `Error`s for user errors. It's slightly ambiguous though in the case of query planning. Supergraph schemas are generated by composition (our own code) and users shouldn't be generating/editing supergraph schemas. But users can technically provide their own supergraph schemas (it's just unsupported), and users may make mistakes using a version of query planning with an incompatible version of composition.
  - For the most part, I've generally followed the pattern in the JS codebase, where JS asserts map to Rust `assert!`/`panic!` and JS places that throw an error with a federation error code map to returning Rust `FederationError`. For JS code that throws a generic error, if there's a federation error code that fits, I go with that. Definitely down to shift more to `FederationError` later, though we'll have to figure out appropriate codes/publish new error codes where needed.
- There are currently no tests; these will be added later.


lrlna (commented):
@sachindshinde I started reviewing this today! Will need a tiny bit more time to wrap it up 🙈

apollo-at-link/src/argument.rs
```diff
@@ -0,0 +1,136 @@
...
+) -> Option<&'directive Name> {
+    application
+        .arguments
+        .iter()
+        .find(|a| *a.name == *name)
+        .and_then(|a| match a.value.deref() {
+            Value::Enum(name) => Some(name),
+            Value::Null => None,
+            _ => {
+                panic!(
```
Geal, line 19: is this used in tests? We generally avoid panicking, because there'll always be something in production tha triggers it. This could return a `Result<Option<&Name>, SomeErrorType>`

apollo-at-link/src/federation_schema.rs
```diff
@@ -0,0 +1,244 @@
...
+impl AsRef<OptionLinksMetadata> for OptionLinksMetadata {
+    fn as_ref(&self) -> &OptionLinksMetadata {
+        self
+    }
+}
+
+// Note that LinksMetadata is not automatically updated for all changes to the schema, and it's up
+// to the caller to determine when re-computation is necessary (in the JS code, this metadata was
+// embedded in the schema as "CoreFeatures").
+pub struct FederationSchema {
```
dariuszkuc, line 34: shouldn't this just map to existing `Subgraph` object instead?

apollo-query-graph/src/extract_subgraphs_from_supergraph.rs
```diff
@@ -0,0 +1,1703 @@
...
+
+// Assumes the given schema has been validated.
+#[allow(dead_code)]
+fn extract_subgraphs_from_supergraph(
+    supergraph_schema: &Schema,
+    validate_extracted_subgraphs: Option<bool>,
+) -> Result<FederationSubgraphs, FederationError> {
+    let validate_extracted_subgraphs = validate_extracted_subgraphs.unwrap_or(true);
+    let (supergraph_schema, link_spec_definition, join_spec_definition) =
+        validate_supergraph(supergraph_schema)?;
```
dariuszkuc, line 36: somewhat curious -> why would we allow passing invalid schema in the first place? maybe this method should be on `Supergraph` to ensure schema is validated?

SimonSapin:
> My general understanding is that panics should be reserved for developer mistakes (e.g. invariant violations in data structures), while we should use `Error`s for user errors.

Yes, this is a good guideline. A panic typically crashes the thread and sometimes the process. It’s technically possible to catch but unusual.

> It's slightly ambiguous though in the case of query planning. Supergraph schemas are generated by composition (our own code) and users shouldn't be generating/editing supergraph schemas.

Even so, the supergraph is an input for the Router and it sometimes looks unexpected. We’ve had cases where in local dev mode, the Router watches the filesystem and reloads when the schema file changes. But if something is writing the file slowly, we ended up trying to load a half-written supergraph schema. IIRC we’ve fixed this particular case (by listening for file-in-write-mode close events, instead of file write events) but in general we want the Router to gracefully handle an invalid supergraph schema.

apollo-query-graph/src/extract_subgraphs_from_supergraph.rs
```diff
@@ -0,0 +1,1703 @@
...
+}
+
+fn collect_empty_subgraphs<'schema, 'a, T: AsRef<OptionLinksMetadata>>(
+    supergraph_schema: &FederationSchemaRef<'schema, T>,
+    join_spec_definition: &'a JoinSpecDefinition,
+) -> Result<
+    (
+        FederationSubgraphs,
+        IndexMap<&'schema Name, &'static FederationSpecDefinition>,
+        IndexMap<&'schema Name, &'schema NodeStr>,
```
lrlna, line 149: I think this can be gathered in a type, and returned as part of this result. It _theoretically_ should help you simplify these lifetimes too.

apollo-at-link/src/spec_definition.rs
```diff
@@ -0,0 +1,157 @@
+use crate::federation_schema::{FederationSchemaRef, OptionLinksMetadata};
+use crate::link::Link;
+use crate::spec::{Identity, Url, Version};
+use apollo_compiler::schema::{DirectiveDefinition, ExtendedType};
+use apollo_compiler::{Node, NodeStr};
+use std::collections::btree_map::Keys;
+use std::collections::BTreeMap;
+use std::sync::Arc;
+
+pub trait SpecDefinition {
```
lrlna, line 10: What's the reasoning for this to be a trait rather than just a struct? From reading, it seems like it could just be a struct and that would help remove some of the generics that get carried along some parts of the code.

apollo-at-link/src/argument.rs
sachindshinde: It's used to extract arguments from directive applications. The expectation here is that the schema should have been validated by this point, so directive applications should have values of the right types (and if they don't, it's the developer's mistake to have not validated).

apollo-at-link/src/link_spec_definition.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn url(&self) -> &Url {
+        &self.url
+    }
+
+    fn minimum_federation_version(&self) -> Option<&Version> {
+        self.minimum_federation_version.as_ref()
+    }
+}
+
+lazy_static! {
```
lrlna, line 33: I am not 100% sure on the current status of `const` for this use case (might already work, or it might be still in the works),  buttttt with recent `const` developments we basically try to use `const` instead of a lot of `lazy_static!` calls we used to do. You may have to refactor this just a bit to get it to work properly with const.

apollo-at-link/src/federation_schema.rs
sachindshinde: `FederationSchema` here just means a schema with federation metadata attached. It could be a subgraph, but it could also be a supergraph (or even a non-link schema, in which case the metadata is `None`).

apollo-at-link/src/join_spec_definition.rs
```diff
@@ -0,0 +1,319 @@
...
+
+pub struct FieldDirectiveArguments<'a> {
+    pub graph: Option<&'a NodeStr>,
+    pub requires: Option<&'a NodeStr>,
+    pub provides: Option<&'a NodeStr>,
+    pub type_: Option<&'a NodeStr>,
+    pub external: Option<bool>,
+    pub override_: Option<&'a NodeStr>,
+    pub user_overridden: Option<bool>,
+}
```
lrlna, line 59: What's the reason for a lifetime for all of these structs to have a lifetime? Is it cause the getters for `NodeStr` in the compiler return a reference? Adding a lifetime to a struct means you'll have to carry it over for all of its uses, which makes a lot of the impls in this PR to have a whole bunch of lifetimes. I am afraid we are going to get ourselves in _a lot_ of trouble with these as we continue on this work.

apollo-at-link/src/join_spec_definition.rs
```diff
@@ -0,0 +1,319 @@
...
+    pub url: &'a NodeStr,
+}
+
+pub struct TypeDirectiveArguments<'a> {
+    pub graph: &'a NodeStr,
+    pub key: Option<&'a NodeStr>,
+    pub extension: bool,
+    pub resolvable: bool,
+    pub is_interface_object: bool,
+}
```
lrlna, line 49: do these have to be `pub` or can they just be `pub(crate)`?

apollo-at-link/src/federation_schema.rs
```diff
@@ -0,0 +1,244 @@
+use crate::database::links_metadata;
+use crate::link::LinksMetadata;
+use crate::referencer::{referencers, Referencers};
+use apollo_compiler::Schema;
+use apollo_federation_error::error::FederationError;
+
+pub mod remove;
+
+// A new type that exists purely to implement AsRef and AsMut on them, as we can't do it on
+// Option<LinksMetadata> directly due to the orphan rule.
```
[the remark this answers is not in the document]
lrlna: I am also of the opinion that we ought to get rid of `AsRef` and `AsMut`. It slightly over complicates a lot of the code, and it looks like only for the reason to have this `LinksMetadata`

apollo-at-link/src/argument.rs
```diff
@@ -0,0 +1,136 @@
+use apollo_compiler::ast::Value;
+use apollo_compiler::schema::{Directive, Name};
+use apollo_compiler::{Node, NodeStr};
+use apollo_federation_error::error::{FederationDirectiveErrorCategory, FederationError};
+use std::ops::Deref;
+
+pub fn directive_optional_enum_argument<'directive>(
+    application: &'directive Node<Directive>,
```
lrlna, line 8: Wondering if these validations(??) can be on a struct itself rather than as single functions. I am hoping this would help us get rid of this lifetime.

apollo-at-link/src/federation_schema/remove.rs
```diff
@@ -0,0 +1,2131 @@
...
+    Component, ComponentStr, Directive, EnumType, EnumValueDefinition, ExtendedType,
+    FieldDefinition, InputObjectType, InputValueDefinition, InterfaceType, Name, ObjectType,
+    ScalarType, UnionType,
+};
+use apollo_compiler::{Node, Schema};
+use indexmap::Equivalent;
+use std::hash::Hash;
+use std::ops::Deref;
+
+impl SchemaDefinitionLocation {
```
lrlna, line 26: I'd have this impl next to the actual definition of `SchemaDefinitionLocation`. We usually like to have impls along with struct definitions.

apollo-at-link/src/federation_schema/remove.rs
```diff
@@ -0,0 +1,2131 @@
...
+    pub fn remove_directive_name<
+        T: AsMut<OptionLinksMetadata> + AsRef<OptionLinksMetadata>,
+        U: AsMut<Referencers> + AsRef<Referencers>,
+        Q: Hash + Equivalent<Name>,
+    >(
+        &self,
+        schema: &mut ReferencerFederationSchemaMut<T, U>,
+        name: &Q,
+    ) {
```
lrlna, line 35: Instead of removing a tuple, I'd either want this to return its own type, or perhaps even the original struct itself without that directive name. It's a bit strange to have it return something vastly different than the original `SchemaDefinitionLocation` itself.

And is the `Location` struct that should be removing a directive? Not federation schema?

lrlna: For a removal, I'd maybe borrow an API that already exists in rust std. For example, when you insert into a hashmap, and a key already exists, it returns the value of the item that was replaced. this could also return the item that was removed. (maybe even also in an option)

apollo-at-link/src/federation_schema/remove.rs
lrlna: This comment is for all of these `impl`s in this file ^

apollo-at-link/src/federation_schema/remove.rs
lrlna: this goes for all of these removal functions in this file ^

apollo-at-link/src/federation_schema.rs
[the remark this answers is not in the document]
sachindshinde: The problem is that in functions, you may have arguments like `&Schema` and `&mut Schema`. And you want to compute metadata for them, like `LinksMetadata` or `Referencers`, and call functions requiring `ReferencerFederationSchemaRef` or `ReferencerFederationSchemaMut`. But if you just have something like
```rust
struct ReferencerFederationSchemaMut {
  schema: &mut Schema,
  metadata: &mut OptionLinksMetadata,
  referencers: &mut Referencers,
}
``` 
Then the struct can't take ownership of `OptionLinksMetadata` and `Referencers`.

You might say that the caller should just compute and take ownership of `OptionLinksMetadata` and `Referencers` and construct a `FederationSchemaMut`, but that violates the abstraction we'd like to present. E.g., we don't want people to have mutable access to `Referencers`; the only things that should mutate `Referencers` are functions in `federation_schema.rs` or its descendants in the module structure. If there's a cleaner/simpler way of maintaining this abstraction we can go with that.

apollo-at-link/src/location.rs
```diff
@@ -0,0 +1,1074 @@
...
+impl From<InputObjectTypeDefinitionLocation> for TypeDefinitionLocation {
+    fn from(value: InputObjectTypeDefinitionLocation) -> Self {
+        TypeDefinitionLocation::InputObjectTypeDefinitionLocation(value)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct SchemaDefinitionLocation;
+
+impl SchemaDefinitionLocation {
```
lrlna, line 57: Why does this need a wrapper struct around the existing `SchemaDefinition`?

apollo-at-link/src/location.rs
```diff
@@ -0,0 +1,1074 @@
...
+}
+
+impl Display for ObjectTypeDefinitionLocation {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.type_name)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ObjectFieldDefinitionLocation {
```
lrlna, line 271: A lot of these seem to be wrappers around what's already in the compiler. What information is missing in the compiler that would make it easier to not have these wrappers/

apollo-query-graph/src/extract_subgraphs_from_supergraph.rs
sachindshinde: FWIW the function is named `validate_supergraph()` in the JS code, but it does very little actual validation on the supergraph schema (it's closer to an `extract_specs()` function than a `validate_supergraph()` function). We don't really have code in the `federation` repo that fully validates a supergraph schema, but rather code that tries to use the supergraph schema to do things (and if there are invalidities discovered along the way, it panics or errors).

We could certainly move this function to `Supergraph` if you'd like, and move the function to a method there as well (and have this `extract_subgraphs_from_supergraph()` function take the `Supergraph` struct instead of `Schema`). I can also rename the function to `extract_specs()` or something. Would that be fine?

lrlna (commented):
I am noticing a couple of things with this PR that could be addressed generally:

1. I'd run clippy on this code. I think it will help a little bit with some of the return type complexity and some of the extra lifetimes. All of our rust projects use clippy just as it is. Very very rarely do we disable a clippy lint for something.
2. It would be really good to think a bit more about lifetimes of some of this functionality. Adding a lifetime to a struct makes that lifetime carry forward to each of its use cases, making code more complex and less maintainable over time. Sometimes it's of course necessary, but a decision that has to be thought about.
3. There are a lot of generic parameters, like `AsRef` and `AsMut` that some of the structs and functions are carrying that might not be necessary. More so than with lifetimes, it's important to really think about whether this is how you want the API to be used. I usually start off with as little generics (and traits, for that matter) as it's possible. I find that a generic-less code is easier to iterate and maintain especially in the beginning.
4. Ditto to what Geoffroy and Simon have said about panics.

sachindshinde:
Pushed up some changes with most of the feedback addressed. Specifically:
- Removed all panics from code introduced in this PR (shouldn't be any new calls to `panic!()`/`assert!()` and `unwrap()`/`expect()`).
- Modified the `error` module to use `thiserror` enums, and to have an `Internal` variant for what used to be panics/asserts.
- Consolidated the `Schema` wrappers into a single one called `FederationSchema`, which now contains no generics/`AsRef`/`AsMut` or lifetime parameters.
- Renamed `*Location` structs to `*Position` structs to avoid name collision, and moved `remove*()` logic to the same file. (Also ended up splitting out referencer computation logic to the same file under `insert*()`.)
- Removed lifetime parameters from structs where I could (e.g. arguments are now cloned, but this should be cheap due to `NodeStr`).
- Fixed Clippy issues (previously it was skipping some files due to some `Cargo.toml` issues), which helped remove a bunch of unnecessary lifetime parameters.
- Fixed some bugs/edge cases along the way (e.g. handling built-ins).

Additionally, I disabled Clippy `dead_code` warnings crate-wide for the time being, as otherwise we'd need to add and subsequently remove several `#[allow(dead_code)]`s across PRs (there's a `TODO` to remove it once we've stabilized more).

Some things I didn't get to:
1. I tried converting `SpecDefinition` to an inner struct instead of a trait, but in `SpecDefinitions` we needed some methods of the trait. I thought about implementing `Deref<SpecDefinition>` but we'd still need to use `T: Deref<SpecDefinition>`, so it doesn't get rid of the generic (not sure what's preferable here).
2. I didn't get time to refactor `extract_subgraphs_from_supergraph()` to carry around its data in a struct, but left a `TODO` here (whenever we get back to implementing Fed 1 supergraph support, we can probably take care of it). Clippy helped remove some unnecessary lifetime parameters here at least.
3. There were some thoughts on using `FederationSchema` as an inner struct of the existing `Subgraph` and `Supergraph` structs from composition, which would be nice, but the refactor would take some time and isn't immediately necessary, so I'm punting it till later.

Let me know if there's anything else y'all think is necessary before merging.

lrlna:
Thank you so much for summarizing all of this! It makes sense to me to push the remaining 3 items for future improvements work.

PR #63, opened 2023-10-23
build: automate release using GH release and circle ci

dariuszkuc:
Trigger CircleCI release job when new GH release (and associated tag) are created

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,7 @@ workflows:
   release:
     jobs:
       - test:
-          name: Test (<< matrix.rust_channel >> rust on << matrix.platform >>)
+          name: Test stable rust on linux
           matrix:
             parameters:
               platform: [linux]
@@ -73,6 +73,8 @@ workflows:
             parameters:
               platform: [linux]
               rust_channel: [stable]
+          requires:
+            - "Test stable rust on linux"
           <<: *crate_release
 
   security-scans:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,14 +24,23 @@ executors:
     resource_class: windows.xlarge
     shell: powershell.exe -ExecutionPolicy Bypass
 
-# There are two workflows: lint and test.
+# Job filter for release jobs
+crate_release: &crate_release
+  filters:
+    branches:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -49,6 +58,23 @@ workflows:
...
+  release:
+    jobs:
+      - test:
+          name: Test (<< matrix.rust_channel >> rust on << matrix.platform >>)
+          matrix:
+            parameters:
+              platform: [linux]
+              rust_channel: [stable]
+          <<: *crate_release
+      - publish:
```
sachindshinde, line 70: We'll want a `requires` key on the `publish` job to depend on the `test` jobs, [similar to federation-rs](https://github.com/apollographql/federation-rs/blob/0757a29c9fb58b2c7f1c9e5ffd9f4eb2a88c6acd/.circleci/config.yml#L101).

.circleci/config.yml
```diff
@@ -102,6 +126,24 @@ jobs:
...
+        default: stable
+      platform:
+        type: executor
+    executor: << parameters.platform >>
+    steps:
+      - checkout
+      - install_system_deps:
+          rust_channel: << parameters.rust_channel >>
+          platform: << parameters.platform >>
+      - run:
```
sachindshinde, line 143: Presumably we'd need to run `cargo login -- $CRATES_IO_API_TOKEN` before doing publish, while putting `CRATES_IO_API_TOKEN` into some new CircleCI context (or if there's already some CircleCI context that the `apollographql` org uses for publishing, we could use that). Or if Rust folks are doing something else to auth with `crates.io`, we can use that.

sachindshinde (changes requested):

.circleci/config.yml
dariuszkuc: I was wondering about this -> I based this on federation-rs where I dont see any explicit login setup. I'm guessing it is already configured on circle ci.

.circleci/config.yml
dariuszkuc: was wondering whether we should just collapse it to a single thing (so we wouldn't have to setup env twice). might do this later

PR #80, opened 2023-11-06
feat: Handle reserved names/meta-fields properly

sachindshinde:
This PR:
- Updates the `federation-next` schema wrapper to handle reserved names better. Specifically, we now:
  - Error if we try to insert/remove elements containing fields/arguments/input fields/enum values that start with `__`.
    - Note that built-in shadowing is still TODO/isn't implemented here, so accordingly we still ignore built-in types/directives as before.
  - Allow fetching meta-fields for objects/interfaces/unions, via `apollo-rs`'s `Schema.type_field()` method.
    - We track their GraphQL references appropriately, similar to the JS code.
    - We also add some convenience functions to get child positions from parent positions, and update code to call them.
  - Rearrange GraphQL reference insertion to occur after `Schema` element insertion instead of before, as we need the type to already be in the schema to call `type_field()` during GraphQL reference insertion. This means doing map lookups post-element-insertion to get a Rust reference to the inserted element, though arguably inserting GraphQL references afterwards is saner behavior as we don't have to worry about the element not existing yet during GraphQL reference insertion.
- Unrelated to the above, we also add some convenience structs/methods used by query graph creation here, namely:
  - `CompositeTypeDefinitionPosition`, `AbstractTypeDefinitionPosition`, and `FieldDefinitionPosition` (mainly just Rust enum analogues to the corresponding TS union types).
  - Some convenience methods for `Referencers` to handle map lookup errors.

### DIFF
--- a/src/schema/position.rs
+++ b/src/schema/position.rs
@@ -20,6 +20,7 @@ use std::hash::Hash;
 use std::ops::Deref;
 use strum::IntoEnumIterator;
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub(crate) enum TypeDefinitionPosition {
     Scalar(ScalarTypeDefinitionPosition),
     Object(ObjectTypeDefinitionPosition),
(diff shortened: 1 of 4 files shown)


src/schema/referencer.rs
```diff
@@ -38,6 +124,7 @@ pub(crate) struct ScalarTypeReferencers {
     pub(crate) object_field_arguments: IndexSet<ObjectFieldArgumentDefinitionPosition>,
     pub(crate) interface_fields: IndexSet<InterfaceFieldDefinitionPosition>,
     pub(crate) interface_field_arguments: IndexSet<InterfaceFieldArgumentDefinitionPosition>,
+    pub(crate) union_fields: IndexSet<UnionTypenameFieldDefinitionPosition>,
```
dariuszkuc, line 127: isn't that union definition that they don't share any fields (as compared to interfaces)?

src/schema/position.rs
```diff
@@ -65,6 +66,123 @@ impl From<InputObjectTypeDefinitionPosition> for TypeDefinitionPosition {
...
+    fn from(value: UnionTypeDefinitionPosition) -> Self {
+        AbstractTypeDefinitionPosition::Union(value)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub(crate) enum FieldDefinitionPosition {
+    Object(ObjectFieldDefinitionPosition),
+    Interface(InterfaceFieldDefinitionPosition),
+    Union(UnionTypenameFieldDefinitionPosition),
```
dariuszkuc, line 144: same question as below -> unions don't have fields so why do we need this? do we track ALL possible fields from all union implementations?

src/schema/position.rs
```diff
@@ -147,6 +266,7 @@ impl SchemaDefinitionPosition {
     fn insert_references(
         &self,
         schema_definition: &Node<SchemaDefinition>,
+        schema: &Schema,
```
dariuszkuc, line 269: since schema contains schema definition do we need both?

src/schema/position.rs
sachindshinde: Unions have a single field named `__typename`, which we need to be able to represent in query graphs.

src/schema/referencer.rs
sachindshinde: See https://github.com/apollographql/federation-next/pull/80#discussion_r1383806697 . There's arguably a question of whether we really need to track this, but the JS code did track dependencies introduced by meta-fields (I haven't checked for usages of it yet though).

src/schema/position.rs
sachindshinde: The second argument is meant to be the element being inserted for all the `insert_references()`-like functions in this file. Sometimes we also need `schema` too, in which case it's also an argument (mostly its used to check the type kind of a type, but in this PR we're also using it for meta-fields via the `apollo-rs` code).

When we have `schema`, we could just omit the 2nd argument as you've said, but we always have the element we're inserting at the callsites, so it saves us some map lookups if we just pass it along as an argument instead of re-fetching.

goto-bus-stop (approved):
I think we can refactor references in the future to require a lot less code, but we can get to that after v0.

PR #84, opened 2023-11-08
feat: Intra-subgraph logic of federated query graph creation

sachindshinde:
This PR implements the intra-subgraph logic of federated query graph creation. That is, it implements the business logic that is performed for each individual subgraph when creating a federated query graph. The inter-subraph logic will be a separate PR, to keep PRs smaller. Tests will be added in follow-up PRs.

Some notes:
- Regarding porting, while the JS codebase creates a separate query graph for each subgraph and then copies its nodes/edges over to the federated one, the Rust code in this PR starts out with a single query graph (the federated one), and each builder adds its nodes/edges to that one. This allows us to avoid copying all the subgraph query graphs (and makes the porting/working with `petgraph` a bit easier).
- We introduce a few convenience functions for `FederationSchema`/positions in this PR. We use those in the `QueryGraph` logic, but I've also refactored `extract_subgraphs_from_supergraph()` logic to use them, which has simplified some functions/removed lifetimes (also some bugfixes).
- This PR is stacked on top of https://github.com/apollographql/federation-next/pull/80 .

### DIFF
--- a/src/query_graph/build_query_graph.rs
+++ b/src/query_graph/build_query_graph.rs
@@ -5,15 +5,18 @@ use crate::link::federation_spec_definition::{
 use crate::link::spec::Identity;
 use crate::link::spec_definition::spec_definitions;
 use crate::query_graph::extract_subgraphs_from_supergraph::extract_subgraphs_from_supergraph;
-use crate::query_graph::{QueryGraph, QueryGraphEdge, QueryGraphEdgeTransition, QueryGraphNode};
+use crate::query_graph::{
+    QueryGraph, QueryGraphEdge, QueryGraphEdgeTransition, QueryGraphNode, QueryGraphNodeType,
+};
 use crate::schema::position::{
     AbstractTypeDefinitionPosition, FieldDefinitionPosition, InterfaceTypeDefinitionPosition,
-    ObjectFieldDefinitionPosition, ObjectTypeDefinitionPosition, SchemaRootDefinitionKind,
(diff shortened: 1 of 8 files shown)


src/query_graph/mod.rs
sachindshinde, line 1: This code defines the query graph data structure and some accessors, corresponding to the analogous parts of [querygraph.ts](https://github.com/apollographql/federation/blob/main/query-graphs-js/src/querygraph.ts) in the JS codebase.

src/query_graph/build_query_graph.rs
sachindshinde, line 1: This code is for building the query graph, corresponding to the relevant parts of [querygraph.ts](https://github.com/apollographql/federation/blob/main/query-graphs-js/src/querygraph.ts). Note that this PR only implements federated query graph creation (API query graph creation is only used by composition and will be deferred to later), and this PR specifically only implements the intra-subgraph part of federated query graph creation.

goto-bus-stop:
> Tests will be added in follow-up PRs.
Are there tests for just the intra-subgraph logic? I think we should aim to get those in as soon as possible, the later we do it the more risk there is of some small discrepancy somewhere in the codebase being super hard to debug.

sachindshinde:
@goto-bus-stop 
Most of the tests are for query planning itself (there's a few tests for query graphs as a whole), there aren't tests directly for intra-subgraph logic from what I can see. It's a division I made primarily to break up a large PR into smaller PRs.

Agreed we should get tests in sooner rather than later; the main incentive behind getting code in is to unblock other work sooner.

PR #85, opened 2023-11-10
First half of updating `federation-next` to use `apollo-compiler@1.0.0-beta.6`

sachindshinde:
This PR exists mainly to unblock the rebasing of #80 and #84, as they'll have merge conflicts with the updated code (and I'd like to update the code for those PRs today). After I've finished rebasing those two PRs, will file a PR for 2nd half (probably later today).

### DIFF
--- a/src/link/federation_spec_definition.rs
+++ b/src/link/federation_spec_definition.rs
@@ -3,22 +3,23 @@ use crate::link::spec::{Identity, Url, Version};
 use crate::link::spec_definition::{SpecDefinition, SpecDefinitions};
 use crate::schema::FederationSchema;
 use apollo_compiler::ast::Argument;
-use apollo_compiler::schema::{Directive, DirectiveDefinition, Value};
-use apollo_compiler::{Node, NodeStr};
+use apollo_compiler::schema::{Directive, DirectiveDefinition, Name, Value};
+use apollo_compiler::{name, Node, NodeStr};
 use lazy_static::lazy_static;
 
-pub(crate) const FEDERATION_KEY_DIRECTIVE_NAME_IN_SPEC: &str = "key";
-pub(crate) const FEDERATION_INTERFACEOBJECT_DIRECTIVE_NAME_IN_SPEC: &str = "interfaceObject";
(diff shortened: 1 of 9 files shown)


dariuszkuc (approved):

src/link/federation_spec_definition.rs
```diff
@@ -3,22 +3,23 @@ use crate::link::spec::{Identity, Url, Version};
...
 use lazy_static::lazy_static;
 
-pub(crate) const FEDERATION_KEY_DIRECTIVE_NAME_IN_SPEC: &str = "key";
-pub(crate) const FEDERATION_INTERFACEOBJECT_DIRECTIVE_NAME_IN_SPEC: &str = "interfaceObject";
-pub(crate) const FEDERATION_EXTERNAL_DIRECTIVE_NAME_IN_SPEC: &str = "external";
-pub(crate) const FEDERATION_REQUIRES_DIRECTIVE_NAME_IN_SPEC: &str = "requires";
-pub(crate) const FEDERATION_PROVIDES_DIRECTIVE_NAME_IN_SPEC: &str = "provides";
-pub(crate) const FEDERATION_SHAREABLE_DIRECTIVE_NAME_IN_SPEC: &str = "shareable";
-pub(crate) const FEDERATION_OVERRIDE_DIRECTIVE_NAME_IN_SPEC: &str = "override";
+pub(crate) static FEDERATION_KEY_DIRECTIVE_NAME_IN_SPEC: Name = name!("key");
```
SimonSapin, line 10: This and other `Name` constants can be `const` items, no need to make them `static`. (The difference is that statics have a fixed address in the memory-mapped executable and can only be accessed by `&` reference.)

SimonSapin (approved):
This already looks nicer than what I had so far! Pre-approved with `static` items changed back to `const`.

src/link/federation_spec_definition.rs
sachindshinde: Chatted with Simon about this in DMs. My initial reasoning for not using `const` here was that I was worried that we would be allocating space for the strings in the `Name`s for each use. It turns out though that `name!()` uses `Name::from_static()`, which avoids the extra allocation and ref-counting for `&'static str`. Will change back to `const`s.

PR #87, opened 2023-11-17
feat: operation normalization for query planner

dariuszkuc:
Initial version of operation normalization logic that auto-expands all fragments, deduplicates the fields and removes top level introspection fields.

related: #88

### DIFF
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -50,16 +50,6 @@ impl NormalizedSelectionSet {
     }
 }
 
-// impl From<SelectionSet> for NormalizedSelectionSet {
-//     fn from(value: SelectionSet) -> Self {
-//         let mut normalized = NormalizedSelectionSet {
-//             ty: value.ty.clone(),
-//             selections: IndexMap::new()
-//         };
-//         normalized
-//     }
-// }
-
 impl From<NormalizedSelectionSet> for SelectionSet {
     fn from(val: NormalizedSelectionSet) -> Self {
         SelectionSet {
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -208,7 +208,7 @@ fn normalize_selections(
                 }
 
                 let mut fragment_key: NormalizedSelectionKey = field.into();
-                // deferred fields should nto be merged
+                // deferred fields should not be merged
                 let is_deferred = is_deferred_selection(&field.directives);
                 if is_deferred {
                     while normalized.contains_key(&fragment_key) {
@@ -389,9 +389,9 @@ fn merge_selections(
                                 // insert new
                                 merged_selections.insert(
                                     fragment_key,
-                                    NormalizedSelection::NormalizedInlineFragment(Node::new(
-                                        fragment_to_merge.deref().clone(),
-                                    )),
+                                    NormalizedSelection::NormalizedInlineFragment(
+                                        fragment_to_merge.clone(),
+                                    ),
                                 );
                             } else {
                                 // can merge
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -350,7 +350,7 @@ type Foo {
     baz
   }
 }"#;
-            let actual = format!("{}", operation);
+            let actual = operation.to_string();
             assert_eq!(expected, actual);
         }
     }
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -3,7 +3,9 @@ use apollo_compiler::executable::{
     Field, Fragment, InlineFragment, Operation, Selection, SelectionSet,
 };
 use apollo_compiler::{Node, Schema};
+use indexmap::map::Entry;
 use indexmap::IndexMap;
+use std::ops::Deref;
 
 // copy of apollo compiler types that store selections in a map so we can normalize it efficiently
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -12,26 +14,49 @@ pub struct NormalizedSelectionSet {
     pub selections: NormalizedSelectionMap,
 }
 
+/// Normalized selection map is an optimized representation of a regular SelectionSet which does not
+/// contains any duplicated entries and optimizes fragment usages. By storing selection set as a map,
+/// we can efficiently join multiple selection sets.
 pub type NormalizedSelectionMap = IndexMap<NormalizedSelectionKey, NormalizedSelection>;
 
+/// Unique identifier of a selection that is used to determine whether fields/fragments can be merged.
+///
+/// In order to merge two selections they need to
+/// * reference the same field/inline fragment
+/// * specify the same directives
+/// * directives have to be applied in the same order
+/// * directive arguments order does not matter (they get automatically sorted by their names).
+/// * selection cannot specify @defer directive
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub enum NormalizedSelectionKey {
     Field {
-        name: Name,
+        // field alias (if specified) or field name in the resulting selection set
+        response_name: Name,
+        // directive applied on the field
         directives: DirectiveList,
+        // unique label/counter used to distinguish fields that cannot be merged
+        label: i32,
     },
     InlineFragment {
+        // optional type condition of a fragment
         type_condition: Option<Name>,
+        // directives applied on a fragment
         directives: DirectiveList,
+        // unique label/counter used to distinguish fragments that cannot be merged
+        label: i32,
     },
 }
 
+// copy of apollo compiler types that store selections in a map so we can normalize it efficiently
+// we no longer have FragmentSpread variant as they get either auto expanded and merged into regular
+// field selection or converted to inline fragments if we cannot merge them
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum NormalizedSelection {
     NormalizedField(Node<NormalizedField>),
     NormalizedInlineFragment(Node<NormalizedInlineFragment>),
 }
 
+// copy of apollo compiler types that store selections in a map so we can normalize it efficiently
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct NormalizedField {
     pub definition: Node<FieldDefinition>,
@@ -42,6 +67,7 @@ pub struct NormalizedField {
     pub selection_set: NormalizedSelectionSet,
 }
 
+// copy of apollo compiler types that store selections in a map so we can normalize it efficiently
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct NormalizedInlineFragment {
     pub type_condition: Option<NamedType>,
@@ -50,13 +76,12 @@ pub struct NormalizedInlineFragment {
 }
 
 impl NormalizedSelectionSet {
-    // cannot use From trait as we need to pass the fragments
-    // otherwise we need two passes - one to change to map and another one to expand fragments
     fn from_selection_set(
         selection_set: &SelectionSet,
         fragments: &IndexMap<Name, Node<Fragment>>,
     ) -> Self {
-        let normalized_selections = normalize_selections(&selection_set.selections, fragments);
+        let normalized_selections =
+            normalize_selections(&selection_set.selections, &selection_set.ty, fragments);
         NormalizedSelectionSet {
             ty: selection_set.ty.clone(),
             selections: normalized_selections,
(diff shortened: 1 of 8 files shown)


dariuszkuc:
note: there are tons of tests still commented out that need some additional functionality, already copied them over from JS side so it was just easier to keep them as is

src/query_plan/operation.rs
```diff
@@ -0,0 +1,426 @@
...
+    Field, Fragment, InlineFragment, Operation, Selection, SelectionSet,
+};
+use apollo_compiler::{Node, Schema};
+use indexmap::IndexMap;
+
+// copy of apollo compiler types that store selections in a map so we can normalize it efficiently
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct NormalizedSelectionSet {
+    pub ty: NamedType,
+    pub selections: IndexMap<String, NormalizedSelection>,
```
SimonSapin, line 12: The meaning of this string as map key isn’t obvious without reading the two `*_key` functions below that create it. PR https://github.com/apollographql/federation-next/pull/95 merges into this PR to use an enum instead.

src/query_plan/operation.rs
```diff
@@ -0,0 +1,426 @@
...
+                    // field_entry.selection_set.selections = merged_selections;
+                }
+            }
+            Selection::FragmentSpread(named_fragment) => {
+                if let Some(fragment) = fragments.get(&named_fragment.fragment_name) {
+                    let expanded_selection_set = normalize_selections(
+                        &fragment.selection_set.selections.to_owned(),
+                        fragments,
+                    );
+                    normalized = merge_selections(&normalized, &expanded_selection_set);
```
SimonSapin, line 108: This normalizes fragment spreads to the normalization of the fragment definition’s selection set. Shouldn’t it create an inline fragment instead, in order to preserve the type condition and directives?

src/query_plan/operation.rs
```diff
@@ -0,0 +1,426 @@
...
+            let operation = operation.make_mut();
+            normalize_operation(operation, &schema, &executable_document.fragments);
+
+            let expected = r#"query NamedFragmentQuery {
+  foo {
+    id
+    bar
+    baz
+  }
+}"#;
```
SimonSapin, line 343: If you find yourself tired of manually updating expected strings when something changes, the https://docs.rs/expect-test/ crate provides a macro so that `UPDATE_EXPECT=1 cargo test` automatically updates them to the new result, which can them be reviewed with `git diff`.

src/query_plan/operation.rs
```diff
@@ -0,0 +1,426 @@
...
+  bar: String!
+  baz: Int
+}
+"#;
+        let (schema, mut executable_document) =
+            apollo_compiler::parse_mixed(operation_with_named_fragment, "document.graphql");
+
+        if let Some(operation) = executable_document
+            .named_operations
+            .get_mut(&Name::new_unchecked(NodeStr::new("NamedFragmentQuery")))
```
SimonSapin, line 332: This should work:
```suggestion
            .named_operations
            .get_mut("NamedFragmentQuery")
```

src/query_plan/operation.rs
```diff
@@ -0,0 +1,426 @@
...
+            normalize_operation(operation, &schema, &executable_document.fragments);
+
+            let expected = r#"query NamedFragmentQuery {
+  foo {
+    id
+    bar
+    baz
+  }
+}"#;
+            let actual = format!("{}", operation);
```
SimonSapin, line 344: Equivalent and maybe a bit easier to type:
```suggestion
            let actual = operation.to_string();
```

src/query_plan/operation.rs
```diff
@@ -0,0 +1,426 @@
...
+
+type Query {
+  foo: String
+}
+"#;
+        let (schema, mut executable_document) =
+            apollo_compiler::parse_mixed(operation_with_introspection, "document.graphql");
+        if let Some(operation) = executable_document
+            .named_operations
+            .get_mut(&Name::new_unchecked(NodeStr::new("TestIntrospectionQuery")))
```
SimonSapin, line 418: Same as above:
```suggestion
            .get_mut("TestIntrospectionQuery")
```

src/query_plan/operation.rs
dariuszkuc: good callout, I'll double check with the JS implementation (but regardless we should still be converting to the inline fragments due to the above)

sachindshinde: @SimonSapin is right here. The relevant code is in [these lines](https://github.com/apollographql/federation/blob/7903a296a96948d96f37e2638a6f3e165a413d28/internals-js/src/operations.ts#L3618-L3620), which gets called by `expandAllFragments()` when dealing with named fragment spreads.

The gist is that if both of the following are true:
1. The named fragment's type equals the parent type of the fragment spread.
2. The named fragment's fragment spread and fragment definition have no directive applications.

Then we can effectively treat the named fragment's selection set's selections as top-level selections, and merge them into `normalized`. Otherwise, we create an inline fragment. (More generally, what's happening here is that the JS codebase is "rebasing" the inline fragment onto its parent as an optimization, since in this case it's cheap computationally and it's guaranteed to succeed).

Two things to note though. The first is that `normalize_selections()` doesn't currently take in any argument for `parent_type`; you'll need to add that argument, and provide it at the callsites from the selection sets (alternatively, it looks like all the current callsites are effectively taking their selections from selection sets, so you could just pass a selection set reference into the function and rename it to e.g. `normalize_selection_set()`).

The second thing is that you might be surprised about point 2, where we oddly look at the fragment definition's directive applications in addition to the fragment spread's. This is part of the JS codebase; if you look at the code [here](https://github.com/apollographql/federation/blob/7903a296a96948d96f37e2638a6f3e165a413d28/internals-js/src/operations.ts#L3472), you'll notice that the directives for what JS codebase calls the "operation element" of the fragment spread consists of the concatenation of the named fragment's definition's directive applications and the spread's directive applications. This accordingly affects the JS codebase's `key` for the fragment spread (but doesn't affect its `toString()`).

I'm tempted to write this off as a bug for now, and to only look at the fragment spread's directive applications. There aren't any spec directives (both in GraphQL spec or the federation spec) that go on the `FRAGMENT_DEFINITION` location. (The other option would be to add a field to `NormalizedSelectionKey::InlineFragment` for storing `named_fragment_directives`, but then if we ever converted the operation to the `apollo-rs` representation or to a string and converted back, that data would be gone.)

src/query_plan/operation.rs
```diff
@@ -0,0 +1,435 @@
...
+    let mut normalized = NormalizedSelectionMap::new();
+    for selection in selections {
+        match selection {
+            Selection::Field(field) => {
+                let expanded_selection_set =
+                    normalize_selections(&field.selection_set.selections.to_owned(), fragments);
+
+                if let NormalizedSelection::NormalizedField(field_entry) = normalized
+                    .entry(field.into())
+                    .or_insert(NormalizedSelection::NormalizedField(Node::new(
```
sachindshinde, line 92: A minor nit, but if you use `.or_insert_with(|| { ... })` here instead of `.or_insert( ... )`, you can avoid the cost of creating the `NormalizedSelection::NormalizedField` when the entry is already present.

src/query_plan/operation.rs
```diff
@@ -0,0 +1,435 @@
...
+                        fragments,
+                    );
+                    normalized = merge_selections(&normalized, &expanded_selection_set);
+                } else {
+                    // no fragment found
+                }
+            }
+            Selection::InlineFragment(inline_fragment) => {
+                let expanded_selection_set = normalize_selections(
+                    &inline_fragment.selection_set.selections.to_owned(),
```
sachindshinde, line 129: Similarly here, you won't need the `to_owned()`.
```suggestion
                    &inline_fragment.selection_set.selections,
```

src/query_plan/operation.rs
```diff
@@ -0,0 +1,435 @@
...
+                    fragments,
+                );
+
+                if let NormalizedSelection::NormalizedInlineFragment(fragment_entry) = normalized
+                    .entry(inline_fragment.into())
+                    .or_insert(NormalizedSelection::NormalizedInlineFragment(Node::new(
```
sachindshinde, line 135: Similar minor nit to above about using `.or_insert_with()`

src/query_plan/operation.rs
```diff
@@ -0,0 +1,435 @@
...
+                        },
+                    )))
+                {
+                    let merged_selections = merge_selections(
+                        &fragment_entry.selection_set.selections,
+                        &expanded_selection_set,
+                    );
+                    let mut selection_set = fragment_entry.selection_set.clone();
+                    selection_set.selections = merged_selections;
+                    fragment_entry.make_mut().selection_set = selection_set;
```
sachindshinde, line 152: Similar to above, I think this can be simplified to
```suggestion
                    fragment_entry.make_mut().selection_set.selections = merged_selections;
```

src/query_plan/operation.rs
```diff
@@ -0,0 +1,435 @@
...
+) -> NormalizedSelectionMap {
+    let mut merged_selections = source.clone();
+    for (key, selection) in to_merge {
+        if source.contains_key(key) {
+            match selection {
+                NormalizedSelection::NormalizedField(field_to_merge) => {
+                    if let Some(NormalizedSelection::NormalizedField(source_field)) =
+                        merged_selections.get_mut(key)
+                    {
+                        // todo skip deferred
```
sachindshinde, line 172: This is for a later PR, but for the `@defer` handling, the gist is that two different `@defer` applications should be viewed as different for the sake of computing keys.

The JS codebase does this by essentially making the key computation spit out a random UUID for a `@defer` directive application. This doesn't work on its own though, since then the "same" directive application would output two different UUIDs if the key were computed twice on it. The JS codebase instead computes the keys for fields/fragment spreads/inline fragments once, and then stores the key in that object (so really the JS codebase is considering different fields/fragment spreads/inline fragments and not different `@defer` applications).

In this case, we could store a random UUID in the `NormalizedSelectionKey` (or use a static counter to generate it). Then, whenever we clone a key or field/fragment spread/inline fragment, we'd have to choose whether it's really the "same" or not (and if not, we change the UUID).

sachindshinde (changes requested):

sachindshinde:
One other thing to add is that we would ideally be returning some kind of `Error` when something unexpected/impossible happens instead of a no-op, as it'll silently mask issues (I've generally been using `SingleFederationError::Internal` and `Result<_, FederationError>`, but whatever makes sense is fine). This kind of error handling is also fine to do in a separate PR though IMO.

src/query_plan/operation.rs
```diff
@@ -0,0 +1,435 @@
...
+            merged_selections.insert(key.to_owned(), selection.clone());
+        }
+    }
+    merged_selections
+}
+
+impl From<&'_ Node<Field>> for NormalizedSelectionKey {
+    fn from(field: &'_ Node<Field>) -> Self {
+        Self::Field {
+            name: field.name.clone(),
```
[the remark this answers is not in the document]
dariuszkuc: yeah wasn't tackling aliases yet (as we then need to auto-alias conflicting fields)
we also need to account for arguments

src/query_plan/operation.rs
```diff
@@ -0,0 +1,431 @@
...
+    _schema: &Schema,
+    fragments: &IndexMap<Name, Node<Fragment>>,
+) {
+    let mut normalized_selection_set =
+        NormalizedSelectionSet::from_selection_set(&operation.selection_set, fragments);
+    // removes top level introspection
+    normalized_selection_set
+        .selections
+        .retain(|key, _| match key {
+            NormalizedSelectionKey::Field { name, .. } => !name.starts_with("__"),
```
SimonSapin, line 293: Shouldn’t this check for `__schema` and `__type` specifically (schema introspection), and let `__typename` through?

dariuszkuc: We only remove those introspection fields from top level selection. Selecting __typename at root should still be handled by router as it useless from query planner perspective.

sachindshinde: Agreed with @dariuszkuc here, the [existing code](https://github.com/apollographql/federation/blob/7903a296a96948d96f37e2638a6f3e165a413d28/query-planner-js/src/buildPlan.ts#L3427) does the same thing (presumably router already handles this edge case with the existing query planner).

src/query_plan/operation.rs
SimonSapin: https://github.com/apollographql/atlas-docs/blob/fed-deep-dive/federation_deepdive.md#query-cleanup says:
> 2) make sure all `@defer` application has a (unique) label (which simplify allows the planner to have an easy way to refer to the applications).

The Router does the same for auth. It does it by modifying every defer directive application:
* If it already has a `label` argument, add an arbitrary prefix to avoid collisions (Router uses `_`)
* If it doesn’t have a `label` argument, give it a unique one (Router uses a simple counter and `u32::to_string`, no randomness)

(This scheme is reverted when processing response data.)

dariuszkuc:
I'll update logic to use results and proper errors in subsequent PR.

sachindshinde (approved):
Just some minor things below, but otherwise should be good to merge after, so pre-emptively approving.

One thing to note is that the normalized key computation in `src/query_plan/operation.rs` handles `@defer` a bit differently than the JS codebase, in ways that might potentially cause issues later (but for now, it's probably fine to keep as is). I mentioned it back [here](https://github.com/apollographql/federation-next/pull/87#discussion_r1401255402), but the gist is that the key is kept constant once the selection (field/inline fragment/named fragment spread) is created, so that even if the selection is rebased, it ends up having the same key.

That's not currently the case for the normalized key in the Rust code, where the value of the key's label is really only constant within the normalized map; if the selection is cloned and added to another selection set (or rebased), the key will change. This may very well not end up mattering, if the JS codebase never really needs to compare keys between two selections that are the same object. It could end up mattering, if e.g. a selection with `@defer` gets cloned into a selection set and we later have to compare them (e.g. for the logic for optimizing subgraph queries to reuse fragments, if the fragments have selections with `@defer`, then I'm not sure how the comparison logic handles that). This is mainly just something to keep an eye out for when porting the remaining operation processing logic.

PR #102, opened 2023-12-06
Use Name instead of strings for names

goto-bus-stop:
while experimenting with diagnostics in the `link` module, I found it useful to have `Name` instances (with location information) to point to.

This also moves some errors earlier--if you for example have a spec URL where the name section is not a valid GraphQL name, it bails out immediately instead of at some later point when trying to use the name.

### DIFF
--- a/src/link/argument.rs
+++ b/src/link/argument.rs
@@ -83,7 +83,7 @@ pub(crate) fn directive_optional_fieldset_argument(
         Some(argument) => match argument.value.deref() {
             Value::String(name) => Ok(Some(name.clone())),
             Value::Null => Ok(None),
-            _ => Err(SingleFederationError::InvalidGraphQL {
+            _ => Err(SingleFederationError::Internal {
                 message: format!("Invalid value for argument \"{}\": must be a string.", name),
             }
             .into()),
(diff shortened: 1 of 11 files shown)


src/link/argument.rs
```diff
@@ -83,7 +83,7 @@ pub(crate) fn directive_optional_fieldset_argument(
         Some(argument) => match argument.value.deref() {
             Value::String(name) => Ok(Some(name.clone())),
             Value::Null => Ok(None),
-            _ => Err(SingleFederationError::InvalidGraphQL {
+            _ => Err(SingleFederationError::Internal {
```
goto-bus-stop, line 86: Turned this into an `Internal` error like the other errors in this file, because it should be caught by validation

src/link/mod.rs
```diff
@@ -116,50 +117,54 @@ impl Import {
...
-                            if !alias_str.starts_with('@') {
-                                Err(LinkError::BootstrapError(format!("invalid alias '{}' for import name '{}': should start with '@' since the imported name does", alias_str, element)))?
-                            }
-                            alias = Some(alias_str.strip_prefix('@').unwrap().to_string());
+                    if let Some(directive_name) = element.strip_prefix('@') {
+                        if let Some(alias_str) = alias.as_ref() {
+                            let Some(alias_str) = alias_str.strip_prefix('@') else {
+                                return Err(LinkError::BootstrapError(format!("invalid alias '{}' for import name '{}': should start with '@' since the imported name does", alias_str, element)));
+                            };
+                            alias = Some(alias_str.into());
```
goto-bus-stop, line 150: The `None`/`Some`-ness of `.strip_prefix()` has the same information as `.starts_with()`. This is a small refactor to only use `.strip_prefix()` instead of two checks + an unwrap

dariuszkuc (approved):
nice!

Issue #30, opened 2023-09-13
Query graph traversal (query plan business logic)


sachindshinde:
During query planning, we need to use the operation to traverse the query graph and generate options for each closed branch (leaf field).

This is specifically step 8 of [this query-planning summary](https://github.com/apollographql/federation-planning/issues/456). Note that if it's possible to cleanly punt on porting feature-specific code for later (e.g. subscriptions), we should take the opportunity to do so, in the interests of delivering incrementally.

hwillson:
Something to consider - can @sachindshinde pair with @SimonSapin on this work, to make sure we're cross-pollinating?

lrlna:
closing as duplicate of smaller issues.

Issue #42, opened 2023-09-18
epic: query planner - query graph traversal (testing)


sachindshinde:
**This ticket has been moved to https://apollographql.atlassian.net/browse/FED-85.**

---

Port the federation repo tests corresponding to the functionality ported during https://github.com/apollographql/federation-next/issues/30

PR #108, opened 2023-12-11
feat: Query Plan nodes

sachindshinde:
This PR adds Rust-centric versions of the data structures for query plan nodes (with an eye toward avoiding unnecessary serdes and using idiomatic Rust where possible).

Part of #30 .

### DIFF
--- a/src/query_plan/mod.rs
+++ b/src/query_plan/mod.rs
@@ -1,5 +1,6 @@
 use apollo_compiler::executable::{Field, InlineFragment, OperationType, Selection, SelectionSet};
-use apollo_compiler::{ExecutableDocument, Node, NodeStr};
+use apollo_compiler::{ExecutableDocument, NodeStr};
+use std::sync::Arc;
 
 pub mod operation;
 pub mod query_planner;
@@ -9,13 +10,13 @@ pub struct QueryPlan {
 }
 
 pub enum TopLevelPlanNode {
-    Subscription(Node<SubscriptionNode>),
-    Fetch(Node<FetchNode>),
-    Sequence(Node<SequenceNode>),
-    Parallel(Node<ParallelNode>),
-    Flatten(Node<FlattenNode>),
-    Defer(Node<DeferNode>),
-    Condition(Node<ConditionNode>),
+    Subscription(SubscriptionNode),
+    Fetch(FetchNode),
+    Sequence(SequenceNode),
+    Parallel(ParallelNode),
+    Flatten(FlattenNode),
+    Defer(DeferNode),
+    Condition(ConditionNode),
 }
 
 pub struct SubscriptionNode {
@@ -24,12 +25,12 @@ pub struct SubscriptionNode {
 }
 
 pub enum PlanNode {
-    Fetch(Node<FetchNode>),
-    Sequence(Node<SequenceNode>),
-    Parallel(Node<ParallelNode>),
-    Flatten(Node<FlattenNode>),
-    Defer(Node<DeferNode>),
-    Condition(Node<ConditionNode>),
+    Fetch(Arc<FetchNode>),
+    Sequence(Arc<SequenceNode>),
+    Parallel(Arc<ParallelNode>),
+    Flatten(Arc<FlattenNode>),
+    Defer(Arc<DeferNode>),
+    Condition(Arc<ConditionNode>),
 }
 
 pub struct FetchNode {
--- a/src/query_plan/mod.rs
+++ b/src/query_plan/mod.rs
@@ -1,2 +1,210 @@
+use apollo_compiler::executable::{Field, InlineFragment, OperationType, Selection, SelectionSet};
+use apollo_compiler::{ExecutableDocument, Node, NodeStr};
+
 pub mod operation;
 pub mod query_planner;
+
+pub struct QueryPlan {
+    node: Option<TopLevelPlanNode>,
+}
+
+pub enum TopLevelPlanNode {
+    Subscription(Node<SubscriptionNode>),
+    Fetch(Node<FetchNode>),
+    Sequence(Node<SequenceNode>),
+    Parallel(Node<ParallelNode>),
+    Flatten(Node<FlattenNode>),
+    Defer(Node<DeferNode>),
+    Condition(Node<ConditionNode>),
+}
+
+pub struct SubscriptionNode {
+    primary: FetchNode,
+    rest: Option<PlanNode>,
+}
+
+pub enum PlanNode {
+    Fetch(Node<FetchNode>),
+    Sequence(Node<SequenceNode>),
+    Parallel(Node<ParallelNode>),
+    Flatten(Node<FlattenNode>),
+    Defer(Node<DeferNode>),
+    Condition(Node<ConditionNode>),
+}
+
+pub struct FetchNode {
+    service_name: NodeStr,
+    /// Optional identifier for the fetch for defer support. All fetches of a given plan will be
+    /// guaranteed to have a unique `id`.
+    id: Option<NodeStr>,
+    /// If query planner defer support is enabled _and_ the subgrpah named `service_name` supports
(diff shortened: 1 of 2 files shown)


src/query_plan/mod.rs
```diff
@@ -1 +1,210 @@
...
+
 pub mod operation;
+pub mod query_planner;
+
+pub struct QueryPlan {
+    node: Option<TopLevelPlanNode>,
+}
+
+pub enum TopLevelPlanNode {
+    Subscription(Node<SubscriptionNode>),
```
goto-bus-stop, line 12: `apollo_compiler::Node` is primarily an `Arc` with source location information, is source location relevant here? If not, this probably should be either an Arc or just nothing (`Subscription(SubscriptionNode)`)

lrlna:
part of #42 or part of #30? 🤔

src/query_plan/mod.rs
sachindshinde: Good callout, there's no relevant source information. Will use `Arc` instead.

sachindshinde: Will also remove `Arc` from `TopLevelPlanNode`.

src/query_plan/mod.rs
```diff
@@ -1 +1,210 @@
...
+
+pub struct FetchNode {
+    service_name: NodeStr,
+    /// Optional identifier for the fetch for defer support. All fetches of a given plan will be
+    /// guaranteed to have a unique `id`.
+    id: Option<NodeStr>,
+    /// If query planner defer support is enabled _and_ the subgrpah named `service_name` supports
+    /// defer, then this boolean says whether `operation` contains some @defer. Unset otherwise.
+    has_defers: Option<bool>,
+    variable_usages: Vec<NodeStr>,
```
lrlna, line 43: should this be a type, just like `requires` below is a `Selection`? I take it this is the variables in arguments, i.e. `Value` type. Although I do understand having it be a `Value` adds a certain level of complexity of working with an enum, when we are only ever interested in the one variant - `Value::Variable`.

lrlna (approved):

src/query_plan/mod.rs
goto-bus-stop: I think if it's only variable names the `Value` enum is a bit much, but it could use `Name` instead of `NodeStr`

src/query_plan/mod.rs
```diff
@@ -1 +1,210 @@
...
+    KeyRenamer(FetchDataKeyRenamer),
+}
+
+/// A rewrite that sets a value at the provided path of the data it is applied to.
+pub struct FetchDataValueSetter {
+    /// Path to the value that is set by this "rewrite".
+    path: Vec<FetchDataPathElement>,
+    /// The value to set at `path`. Note that the query planner currently only uses string values,
+    /// but that may change in the future.
+    set_value_to: serde_json::value::Value,
```
goto-bus-stop, line 173:
```suggestion
    set_value_to: serde_json::Value,
```
Ultra nitpick but generally using the top-level reexport is preferred

src/query_plan/mod.rs
```diff
@@ -1 +1,210 @@
...
+    node: Option<PlanNode>,
+}
+
+/// A deferred block of a `DeferNode`.
+pub struct DeferredDeferBlock {
+    /// References one or more fetch node(s) (by `id`) within `DeferNode.primary.node`. The plan of
+    /// this deferred part should not be started until all such fetches return.
+    depends: Vec<DeferredDependency>,
+    /// The optional defer label.
+    label: Option<NodeStr>,
```
SimonSapin, line 127: Is this really optional? Isn’t there code to add labels to `@defer` directive applications that don’t already have one?

src/query_plan/mod.rs
```diff
@@ -1 +1,210 @@
...
+    /// Path, in the query, to the `@defer` application this corresponds to. The `sub_selection`
+    /// starts at this `query_path`.
```
SimonSapin, line 129: “`query_path` starts at this `query_path`” doesn’t sound right. Should it be something like “this `query_path` starts at the `query_path` of the nearest `DeferredDeferBlock` ancestor node, if any”?

src/query_plan/mod.rs
```diff
@@ -1 +1,210 @@
...
+    /// A `FetchNode` ID.
+    id: NodeStr,
+    /// If the `FetchNode` pointed to by `id` has `has_defers` as `true`, and this field is set (to
+    /// the label of one of the defers of said `FetchNode`), then this dependency is strictly for
+    /// the defers with that label.
+    defer_label: Option<NodeStr>,
+}
+
+pub struct ConditionNode {
+    condition: NodeStr,
```
SimonSapin, line 153: For doc-comments: what does this string contain? Is it the name of a boolean variable? An arbitrary expression in some language?

src/query_plan/mod.rs
sachindshinde: It's indeed just variable names, changing it to `Name`s should be better.

src/query_plan/mod.rs
sachindshinde: Good catch!

src/query_plan/mod.rs
```diff
@@ -1 +1,210 @@
...
+    /// `Selection`s in apollo-rs _can_ have a `FragmentSpread`, but this `Selection` is
+    /// specifically typing the `requires` key in a built query plan, where there can't be
+    /// `FragmentSpread`.
+    // PORT_NOTE: This was its own type in the JS codebase, but it's likely simpler to just have the
+    // constraint be implicit for router instead of creating a new type.
+    requires: Vec<Selection>,
+    // PORT_NOTE: We don't serialize the "operation" string in this struct, as these query plan
+    // nodes are meant for direct consumption by router (without any serdes), so we leave the
+    // question of whether it needs to be serialized to router.
+    operation_document: ExecutableDocument,
```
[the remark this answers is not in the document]
sachindshinde: > After a document is put inside a fetch node, is it ever mutated again?

I'm not sure about whether router typically mutates this document (or whether it will in the near future). @Geal may know more here.

> Is it expected to always be valid? If no and yes respectively, this should be Valid<ExecutableDocument>.

It should be valid (though I don't think the JS codebase ever fully validates it; at best, since the JS codebase uses a representation with some built-in validations, we get those, but not the rest). Even if we're not sure about mutation, I'd say it's fine to use `Valid<ExecutableDocument>` here with `assume_valid` with a way to optionally do full validation for debugging.

> If we consider it a bug if the query planner emits invalid document, the code creating fetch nodes could validate only when debug assertions are enabled, like in https://github.com/apollographql/apollo-rs/blob/d97abbc94299c998e7e8eac762adc315917da127/crates/apollo-compiler/src/execution/introspection_split.rs#L240-L246

I was gonna suggest passing down a variable like we do for `extract_subgraphs_from_supergraph()`, but using `debug_assertions` looks better. Thanks for the tip!

src/query_plan/mod.rs
sachindshinde: The defer normalization within query planning makes sure that every `@defer` application without a label has a unique label. However, we explicitly do not include such assigned labels in the query plan ([relevant code here](https://github.com/apollographql/federation/blob/f69a0694b95e99c22f0d5f7bc6df1b46568b2a8d/query-planner-js/src/buildPlan.ts#L3607)). FWIW I'm not sure this is bug-free code, but this is currently the code that's there.

src/query_plan/mod.rs
sachindshinde: If I understand correctly, what it's saying is that the `sub_selection` (the next field in this struct) is a selection set whose root is at the element pointed to by this `query_path`.

src/query_plan/mod.rs
sachindshinde: It's currently only the name of a boolean variable (although that may change in the future). I'll rename it to `condition_variable` and make its type `Name` to make it clearer, and then later if we need to generalize we can add a generic `condition` with an appropriate type.

Issue #104, opened 2023-12-07
[Operation Processing] Add the ability to represent normalized named fragments and named fragment spreads


sachindshinde:
The normalized representation of operations introduced in #87 cannot currently represent named fragment spreads (e.g. `... on Foo` for named fragment `Foo`) and named fragments. This is fine for now, since we needed to expand fragments anyway prior to query planning. However:
1. When we get to fragment optimization of subgraph queries after query planning, we'll need to represent subgraph queries with named fragment spreads (and their associated named fragments).
2. In order to facilitate the key logic in #103, we'll need to be able to convert `apollo-rs` named fragments to normalized named fragments first, and then expand.

This ticket is for adding the ability to represent normalized named fragments, along with named fragment spreads in normalized selection sets.

Issue #103, opened 2023-12-07
[Operation Processing] Compute normalized keys in operations per-element instead of per-normalized-map


sachindshinde:
When we normalize `apollo-rs` operations in #87 and convert them to our normalized representation, we compute a key for each selection in a selection set. In the JS code, this key is computed once and stored in the selection's element itself (i.e. the field/inline fragment/named fragment spread). This allows it to stay the same when that element is copied, such as when a named fragment is expanded, or when selections are copied from a supergraph query into a subgraph query.

However, the Rust code currently computes this when the `NormalizedMap` is populated during creation. This means that copying the selection (or even moving it around) will change its key. This may have been fine if the key were only used here, but it turns out [we check for key equivalence when we try to reuse fragments in subgraph queries](https://github.com/apollographql/federation/blob/160299f5db19003514ea60a2ffcb8c0b3b001076/internals-js/src/operations.ts#L1869). For this code to work, we'll need to do a few things:

1. The different kinds of selections (fields/inline fragments/named fragment spreads) need a field for storing a `u64` ID.
2. When the selection is created (either from the `apollo-rs` representation, or programmatically via e.g. `new()`), we set the ID from a global incrementing counter (starting at `1`).
3. Selection types need a `key()` method, which computes and returns the current `NormalizedKey`, but the `label` is `0` if there are no `@defer` directive applications, and the above ID otherwise.
4. Cloning should copy this ID. There could optionally be a method to change the ID to a new one taken from the global incrementing counter, if needed (e.g. to change the identity of a `@defer` usage post-clone).

Note that for this to work, we'll also need to do two other things:

1. We'll need to use the same normalized representation throughout query planning (we can't inter-convert with the `apollo-rs` representation).
6. We'll need to ensure named fragments are converted to a normalized form before they are expanded within an operation, so that `@defer`ed selections within them are assigned the same ID (currently `apollo-rs` named fragments are expanded directly into inline fragments). This depends on #104.

sachindshinde:
For the global incrementing counter, @goto-bus-stop has recommended something similar to
```rust
use std::sync::atomic;
// Global storage for the counter
static NEXT: atomic::AtomicUsize = atomic::AtomicUsize::new(0);

// opaque newtype for the ID
#[derive(PartialEq, Eq)]
struct MyGlobalId(usize);
impl MyGlobalId {
    fn new() -> Self {
        // atomically increment global counter
[...]

PR #110, opened 2023-12-11
[op normalization] add NormalizedFragmentSpread and NormalizedFragment

dariuszkuc:
related: https://github.com/apollographql/federation-next/issues/104

### DIFF
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -2,7 +2,7 @@ use crate::error::FederationError;
 use crate::error::SingleFederationError::Internal;
 use apollo_compiler::ast::{Argument, DirectiveList, FieldDefinition, Name, NamedType};
 use apollo_compiler::executable::{
-    Field, Fragment, InlineFragment, Operation, Selection, SelectionSet,
+    Field, Fragment, FragmentSpread, InlineFragment, Operation, Selection, SelectionSet,
 };
 use apollo_compiler::{Node, Schema};
 use indexmap::map::Entry;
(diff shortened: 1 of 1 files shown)


src/query_plan/operation.rs
```diff
@@ -31,7 +31,15 @@ pub enum NormalizedSelectionKey {
     Field {
         // field alias (if specified) or field name in the resulting selection set
         response_name: Name,
-        // directive applied on the field
+        // directives applied on the field
+        directives: DirectiveList,
+        // unique label/counter used to distinguish fields that cannot be merged
+        label: i32,
```
lrlna, line 37: is an i32 enough? not exactly sure how we use this, but do we think we need an atomic counter here? or will that be an overkill for its usecase?

SimonSapin: This is only unique within one document, right? I’d expect any document to reach the parser’s token limit well before it contains two billion selections.

src/query_plan/operation.rs
```diff
@@ -406,6 +437,9 @@ fn merge_selections(
                             }
                         }
                     }
+                    NormalizedSelection::NormalizedFragmentSpread(_) => {
+                        // nothing to do fragment spread is already part of the selection set
```
sachindshinde, line 455: nit: Might be nice to copy over [the relevant comment from Sylvain here](https://github.com/apollographql/federation/blob/f69a0694b95e99c22f0d5f7bc6df1b46568b2a8d/internals-js/src/operations.ts#L2182-L2185).

src/query_plan/operation.rs
sachindshinde: For now it's unique within the document. That will change once https://github.com/apollographql/federation-next/issues/103 is implemented (but I think the switch to `i64` can be put off till then).

sachindshinde (approved):

src/query_plan/operation.rs
lrlna: ^ sgtm

PR #20, opened 2023-08-19
chore(deps): update rust crate env_logger to 0.10

svc-secops:
This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| [env_logger](https://togithub.com/rust-cli/env_logger) | dev-dependencies | minor | `0.7` -> `0.10` |

---

### Release Notes

<details>
<summary>rust-cli/env_logger (env_logger)</summary>

### [`v0.10.1`](https://togithub.com/rust-cli/env_logger/blob/HEAD/CHANGELOG.md#0101---2023-11-10)

[Compare Source](https://togithub.com/rust-cli/env_logger/compare/v0.10.0...v0.10.1)

### [`v0.10.0`](https://togithub.com/rust-cli/env_logger/blob/HEAD/CHANGELOG.md#0100---2022-11-24)

[Compare Source](https://togithub.com/rust-cli/env_logger/compare/v0.9.3...v0.10.0)

MSRV changed to 1.60 to hide optional dependencies

##### Fixes

-   Resolved soundness issue by switching from `atty` to `is-terminal`

##### Breaking Changes

To open room for changing dependencies:

-   Renamed `termcolor` feature to `color`
-   Renamed `atty` feature to `auto-color`

### [`v0.9.3`](https://togithub.com/rust-cli/env_logger/blob/HEAD/CHANGELOG.md#093---2022-11-07)

[Compare Source](https://togithub.com/rust-cli/env_logger/compare/v0.9.2...v0.9.3)

-   Fix a regression from v0.9.2 where env_logger would fail to compile with the termcolor feature turned off.

### [`v0.9.2`](https://togithub.com/rust-cli/env_logger/blob/HEAD/CHANGELOG.md#092---2022-11-07)

[Compare Source](https://togithub.com/rust-cli/env_logger/compare/v0.9.1...v0.9.2)

-   Fix and un-deprecate Target::Pipe, which was basically not working at all before and deprecated in 0.9.1.

### [`v0.9.1`](https://togithub.com/rust-cli/env_logger/compare/v0.9.0...v0.9.1)

[Compare Source](https://togithub.com/rust-cli/env_logger/compare/v0.9.0...v0.9.1)

### [`v0.9.0`](https://togithub.com/rust-cli/env_logger/blob/HEAD/CHANGELOG.md#090----2022-07-14)

[Compare Source](https://togithub.com/rust-cli/env_logger/compare/v0.8.4...v0.9.0)

##### Breaking Changes

[...]

PR #111, opened 2023-12-12
chore(deps): remove `env_logger`

goto-bus-stop:
It's not currently used anywhere. alternative to #20 :)

Or **should** we be using it somewhere?


lrlna (approved):

sachindshinde (approved):
It'll be used when traversing the query graph, but arguably that PR can just add it back (alternatively, if there's a better logger we should use, lemme know!)

goto-bus-stop:
got it. env_logger is perfectly fine, we also use it in apollo-rs fuzz tests.

Issue #77, opened 2023-11-03
Compute best fetch dependency graph given the options for each branch (query plan business logic)


sachindshinde:
This GH issue is the first part of https://github.com/apollographql/federation-next/issues/31 .

Specifically:
- Pruning the options for each branch.
- Generating the initial partial fetch dependency graph and remaining options.
- Iterating through all possible fetch dependency graphs from the initial one, and returning the one with best cost.

More specifically, this issue is for porting `QueryPlanningTraversal.computeBestPlanFromClosedBranches()` and its code dependencies.

hwillson:
Something to consider - can @sachindshinde pair with @SimonSapin on this work, to make sure we're cross-pollinating?

abernix:
@sachindshinde can you drop links to the relevant JS code? thx!

lrlna:
closing as duplicate of smaller issues.

PR #112, opened 2023-12-14
feat: Add structs for `GraphPath`, `PathTree`, and `FetchDependencyGraph`

sachindshinde:
This PR adds data structures and doc strings for:
- `GraphPath`
  - Represents a path within a query graph.
- `PathTree` 
  - Represents a collection of paths, condensed into a tree.
- `FetchDependencyGraph`
  - Represents a graph of fetches/fetch groups, where there's an edge from one fetch to another if the tail depends on the head.
- `QueryPlanningTraversal`
  - Represents the state and configuration that's maintained while generating a query plan.
 
This PR also adds various other dependent and consequent data structures.

Part of #30.


src/query_graph/graph_path.rs
```diff
@@ -0,0 +1,228 @@
...
+// in the Rust code we don't have a distinguished type for that case. We instead check this at
+// runtime (introducing the new field `head_must_be_root`). This means the `RootPath` type in the
+// JS codebase is replaced with this one.
+#[derive(Debug, Clone)]
+pub(crate) struct GraphPath<TTrigger: Eq + Hash, TEdge: Into<Option<EdgeIndex>>> {
+    /// The query graph of which this is a path.
+    graph: Arc<QueryGraph>,
+    /// The node at which the path starts (the head node of the first edge in path, aliased here for
+    /// convenience).
+    head: NodeIndex,
```
SimonSapin, line 64: Rather than have duplicated data that could accidentally get out of sync, would a method provide the same convenience? This same info is in `self.edges.first()`, right?

(Same for `tail` and `edge_to_tail`)

src/query_graph/graph_path.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+/// Records, as we walk a path within a GraphQL operation, important directives encountered
+/// (currently `@include` and `@skip` with their conditions).
+#[derive(Debug, Clone, Eq)]
+pub(crate) struct OpGraphPathContext {
+    /// A list of conditionals (e.g. `[{ kind: Include, value: true}, { kind: Skip, value: $foo }]`)
+    /// in the reverse order in which they were applied (so the first element is the inner-most
+    /// applied include/skip).
+    conditionals: Vec<Arc<OperationConditional>>,
```
SimonSapin, line 196: Can we ever have more than two of these? Or could the Vec be replaced with `include: Option<BooleanOrVariable>, skip: Option<BooleanOrVariable>`?

src/query_graph/graph_path.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+impl PartialEq for OpGraphPathContext {
+    fn eq(&self, _other: &Self) -> bool {
+        todo!()
```
SimonSapin, line 201: Is the intent to replace this before merging?

SimonSapin (approved):

src/query_graph/graph_path.rs
sachindshinde: I think the `head`/`tail` actually point to the first/last non-null edge in the list (and if no such edges exist, `head`/`tail` point to the same node). I've updated the doc strings for them to be more clear at least. I think we can eliminate `edge_to_tail` (in the JS codebase it was eliminating two layers of indirection while here in Rust it's only eliminating one). Similarly, I'll also remove `SubgraphEnteringEdgeInfo.edge`.

src/query_graph/graph_path.rs
sachindshinde: Yeah, there can be a variable number (the example just happens to have two).

src/query_graph/graph_path.rs
sachindshinde: Nah, a separate PR I file will fill in the implementation (this PR is mainly to unblock work that overlaps a bit).

sachindshinde:
Note that I ended up adding in a struct definition for `QueryPlanningTraversal` to unblock some work this week on #77. (I also tweaked operations to use `LinkedHashMap` to ensure removals don't change order.)

PR #113, opened 2023-12-15
[op normalization] optimize sibling typename usage

dariuszkuc:
Initial implementation of logic to optimize usage of typenames in the operation.

see: https://github.com/apollographql/federation/blob/main/query-planner-js/src/buildPlan.ts#L3192

resolves: https://github.com/apollographql/federation-next/issues/89


src/query_plan/operation.rs
```diff
@@ -473,6 +479,106 @@ fn is_deferred_selection(directives: &DirectiveList) -> bool {
...
+                                .alias
+                                .clone()
+                                .unwrap_or_else(|| typename_field.name.clone())
+                                .to_string(),
+                        );
+                    }
+                }
+            }
+        }
+    }
```
goto-bus-stop, line 578: Here you end up checking if `typename_field_key` and `sibling_field_key` are `Some()` twice. It could be flattened a bit by pattern matching on a tuple:

```suggestion
    if let (Some(typename_key), Some(sibling_field_key)) = (typename_field_key, sibling_field_key) {
        if let (
            Some(NormalizedSelection::NormalizedField(typename_field)),
            Some(NormalizedSelection::NormalizedField(sibling_field)),
        ) = (
            selections.shift_remove(&typename_key),
            selections.get_mut(&sibling_field_key),
        ) {
            sibling_field.make_mut().attachments.insert(
                SIBLING_TYPENAME_KEY.to_string(),
                typename_field
                    .alias
                    .clone()
                    .unwrap_or_else(|| typename_field.name.clone())
                    .to_string(),
            );
        } else {
            unreachable!("typename and sibling fields must both exist at this point")
        }
    }
```
I'd also add an `unreachable!()` assertion. The code before this guarantees that the `typename_field_key` and `sibling_field_key` variables point to the correct thing, but we shouldn't silently ignore it if a bug is introduced there in the future.

src/query_plan/operation.rs
```diff
@@ -473,6 +479,106 @@ fn is_deferred_selection(directives: &DirectiveList) -> bool {
...
+/// that case, and that's actually what we want.
+fn optimize_sibling_typenames(
+    normalized_selection_set: &mut NormalizedSelectionSet,
+    interface_types_with_interface_objects: &IndexSet<InterfaceTypeDefinitionPosition>,
+) -> Result<(), FederationError> {
+    let selections = &mut normalized_selection_set.selections;
+
+    let is_interface_object = interface_types_with_interface_objects
+        .iter()
+        .any(|intf| intf.type_name == normalized_selection_set.ty);
```
goto-bus-stop, line 520: Since this is a set, we could do something like this to make it O(1). Very minor knowing it's likely not going to be the most expensive thing in the world anyway.
```rust
let is_interface_object = interface_types_with_interface_objects.contains(&InterfaceTypeDefinitionPosition {
    type_name: normalized_selection_set.ty.clone(),
});
```

goto-bus-stop (commented):
Looks good, got a few stylistic/Rust comments below!

src/query_plan/operation.rs
dariuszkuc: Oo TIL we can pattern match tuples! 🚀 ❤️

dariuszkuc:
For simplicity, instead of using generic `attachment` map we are now using an option field instead.

Since `normalize_operation` is only public within a crate, I moved integration tests back to sources as unit tests.

sachindshinde (approved):
LGTM!

Issue #105, opened 2023-12-07
[Plan Selection] Computing best plan


sachindshinde:
Port the logic for `generateAllPlansAndFindBest()` from [generateAllPlans.ts](https://github.com/apollographql/federation/blob/160299f5db19003514ea60a2ffcb8c0b3b001076/query-planner-js/src/generateAllPlans.ts) in the JS codebase over to Rust. The logic is relatively self-contained, so it's a candidate for a good first issue in the sense it doesn't require much prerequisite reading. The testing is also pretty self-contained (and there's not too much test logic), so it shouldn't be too bad to port tests either.

Part of #77.

PR #114, opened 2023-12-15
[Plan Selection] Computing best plan

SimonSapin:
Fixes https://github.com/apollographql/federation-next/issues/105

Ports logic and unit tests from:
* https://github.com/apollographql/federation/blob/f69a0694b/query-planner-js/src/generateAllPlans.ts
* https://github.com/apollographql/federation/blob/f69a0694b/query-planner-js/src/__tests__/generateAllPlans.test.ts


src/query_plan/generate.rs
```diff
@@ -0,0 +1,317 @@
+use std::collections::VecDeque;
+
+/// TODO: should it be an integer?
+type Cost = f64;
```
sachindshinde, line 4: We have a declaration for a cost type in `query_plan/mod.rs`
```rust
pub type QueryPlanCost = i64;
```

sachindshinde (approved):
Looks mostly good, just one nit below.

src/query_plan/generate.rs
```diff
@@ -0,0 +1,317 @@
...
+///   we can use to plan that particular element.
+/// * `add_function`: how to obtain a new plan by taking some plan and adding a new element to it.
+/// * `cost_function`: how to compute the cost of a plan.
+/// * `on_plan`: an optional method called on every _complete_ plan generated by this method,
+///   with both the cost of that plan and the best cost we have generated thus far
+///   (if that's not the first plan generated).
+///   This mostly exists to allow some debugging.
+pub fn generate_all_plans_and_find_best<Plan, Element>(
+    initial: Plan,
+    to_add: Vec<Choices<Element>>,
```
sachindshinde, line 86: One thing to note for `to_add`: its type here is `Vec<Vec<Option<T>>>`, but it doesn't really make sense to pass in a `None` here (that's a value used by this function to indicate a choice has been searched), so it would be nice to either:
1) Have it be `Vec<Vec<T>>` and convert it in the function, or
2) Update the docstring to indicate callers shouldn't pass in `None`.

SimonSapin: I’ve updated the doc comment, but either way a raw `Vec<Vec<_>>` (as opposed to `Vec` of some struct that contains `Vec`) feels a bit unusual. What do we expect the data structures to look like at the non-test call sites? Could we adapt the function so that no potentially-expensive conversion is needed?

src/query_plan/generate.rs
```diff
@@ -0,0 +1,321 @@
...
+///
+/// Note that this method abstracts the actual types of both plans
+/// and additional elements to add to the plan,
+/// and this both for clarity and to make testing of this method easier.
+/// But type parameter `Plan` should be though of as abstracting a query plan
+/// (in practice, it is instanciated to a pair of a (`DependencyGraph`, corresponding `PathTree`)),
+/// whith `Eelement` should be though of as an additional element
+/// to add to the plan to make it complete
+/// (instanciated in practice by a `PathTree` for ... reasons ...
+/// but one that really correspond to a single `GraphPath`).
```
dariuszkuc, line 32: fixing some typos

```suggestion
/// Note that this method abstracts the actual types of both plans (type parameter `Plan`) and 
/// additional elements (type parameter `Element`) to add to the plan. This is done for both the 
/// clarity and to make testing of this method easier. 
///
/// Type parameter `Plan` should be thought of as abstracting a query plan but in practice, it is 
/// instantiated to a pair of a (`DependencyGraph`, corresponding `PathTree`).
/// 
/// Type parameter `Element` should be thought of as an additional element to add to the plan to 
/// make it complete. It instantiated in practice by a `PathTree` (for ... reasons ...) but one 
/// that really correspond to a single `GraphPath`.
```

src/query_plan/generate.rs
```diff
@@ -0,0 +1,321 @@
...
+    initial: Plan,
+    to_add: Vec<Choices<Element>>,
+    mut add_function: impl FnMut(&Plan, Element) -> Plan,
+    mut cost_function: impl FnMut(&Plan) -> QueryPlanCost,
+    mut on_plan: impl FnMut(&Plan, QueryPlanCost, Option<QueryPlanCost>),
+) -> (Plan, QueryPlanCost)
+where
+    Element: Clone,
+    // Uncomment to use `dbg!()`
+    // Element: std::fmt::Debug,
```
dariuszkuc, line 94: should this be deleted?

src/query_plan/generate.rs
```diff
@@ -0,0 +1,321 @@
...
+    let mut stack = VecDeque::new();
+    stack.push_back(Partial {
+        partial_plan: initial,
+        partial_cost: None,
+        remaining: to_add.into_iter(),
+        is_root: true,
+        index: Some(0),
+    });
+
+    let mut min = None;
```
dariuszkuc, line 118: wondering whether using `Option` here is better than just initializing it to max i64?

dariuszkuc: guess we could potentially generate with max i64 cost... but that seems unlikely (guess never say never)?

SimonSapin: `Option` matches the TypeScript version: `let min: { best: P, cost: number } | undefined = undefined;`. Is there a reason to prefer something else?

src/query_plan/generate.rs
SimonSapin: I left it intentionally in case someone later wants to use `dbg!` again to print-debug this function, but I don’t mind removing it.

src/query_plan/generate.rs
dariuszkuc: my bad I missed that it is a tuple vs just cost

src/query_plan/generate.rs
SimonSapin: I reformatted your edits to preserve [semantic line breaks](https://sembr.org/) and make the diff easier to read.

src/query_plan/generate.rs
sachindshinde: @SimonSapin 
In the JS codebase, the part that generates `to_add` looks like:
```typescript
const otherTrees = this
  .closedBranches
  .slice(0, idxFirstOfLengthOne)
  .map(b => b.map(opt => PathTree.createFromOpPaths(federatedQueryGraph, root, flattenClosedPath(opt))));
```
The map-within-a-map would presumably create the `Vec<Vec<_>>` (I imagine we could just put a `Some()` there to get the `Option<T>`, and leave a comment about why).

Issue #116, opened 2023-12-18
[Plan Selection] Option pruning for closed branches


sachindshinde:
Port the logic for options pruning from the JS codebase to Rust. That is, port these methods on `QueryPlanningTraversal`:
- [pruneClosedBranches()](https://github.com/apollographql/federation/blob/f69a0694b95e99c22f0d5f7bc6df1b46568b2a8d/query-planner-js/src/buildPlan.ts#L610)
- [sortOptionsInClosedBranches()](https://github.com/apollographql/federation/blob/f69a0694b95e99c22f0d5f7bc6df1b46568b2a8d/query-planner-js/src/buildPlan.ts#L639)
- [This block of code](https://github.com/apollographql/federation/blob/f69a0694b95e99c22f0d5f7bc6df1b46568b2a8d/query-planner-js/src/buildPlan.ts#L683-L700), which should really be moved into a new method named e.g. `reduce_options_if_needed()`.

It'll help to look at the comments in [this section of code](https://github.com/apollographql/federation/blob/f69a0694b95e99c22f0d5f7bc6df1b46568b2a8d/query-planner-js/src/buildPlan.ts#L652-L681), which explains more about those functions in the context in which they're called.

Part of #77.

PR #117, opened 2023-12-18
[Plan Selection] Option pruning for closed branches

SimonSapin:
Fixes https://github.com/apollographql/federation-next/issues/116


src/query_plan/query_planning_traversal.rs
```diff
@@ -86,3 +86,173 @@ struct BestQueryPlanInfo {
...
+
+        // FIXME: doesn’t this leave the element foremly at `i - 1` out of place?
+        // Shouldn’t the whole range be shifted?
+        //
+        // It may be easier to do a full sort again, since it’s documented as:
+        //
+        // https://doc.rust-lang.org/std/primitive.slice.html#method.sort_by
+        // > The current algorithm is an adaptive, iterative merge sort inspired by timsort.
+        // > It is designed to be very fast in cases where the slice is nearly sorted
+    }
```
SimonSapin, line 257: It looks like the TypeScript code has a bug (compared to what comment say the intent is), although this only affects the heuristic reducing options semi-arbitrarily.

src/query_plan/query_planning_traversal.rs
```diff
@@ -86,3 +87,181 @@ struct BestQueryPlanInfo {
...
+                break;
+            }
+            // we remove the right-most option of the first branch,
+            // and them move that branch to it's new place.
+            let prev_size = first_branch.0.len();
+            first_branch.0.pop();
+            plan_count -= plan_count / prev_size;
+            self.reorder_first_branch();
+            // Note that if firstBranch is our only branch, it's fine,
+            // we'll continue to remove options from it (but that is beyond unlikely).
```
sachindshinde, line 235: This comment should be moved to the line that sets `first_branch`.

src/query_plan/query_planning_traversal.rs
sachindshinde: The logic in the JS codebase turns out to be correct, but the JS codebase comments are unfortunately poor/don't describe why.

The key thing the comments don't point out is that the first branch's size only ever decrements by 1. This means that when restoring order, we'll see the first branch (let's say of size `n`), and then potentially a run of elements with size `n+1`, and then any elements after that are size of at most `n`. The `i-1` in the code is the last element of this "run" of `n+1`-sized elements (or the first branch itself, if the run doesn't exist). This means the swap will place an `n+1`-sized element first (which should be fine, since the run of elements after is `n+1`-sized), and then the first branch of size `n` will sit after the `n+1`-sized run but before an element of size `<=n`. So the `Vec` effectively remains sorted.

The gist is it should be fine to remove this `FIXME` comment, but it would help to explain this requirement/behavior in the docstring of `reorder_first_branch()`.

sachindshinde (changes requested):

src/query_plan/query_planning_traversal.rs
```diff
@@ -86,3 +87,181 @@ struct BestQueryPlanInfo {
...
+    ///    and putting those more-likely-to-be-good options first helps this.
+    fn sort_options_in_closed_branches(&mut self) -> Result<(), FederationError> {
+        for branch in &mut self.closed_branches {
+            let mut result = Ok(());
+            branch.0.sort_by_key(|closed_path| {
+                closed_path
+                    .paths
+                    .0
+                    .iter()
+                    .try_fold(0, |sum, path| Ok(sum + path.subgraph_jumps()?))
```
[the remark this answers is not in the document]
SimonSapin: Fixed

src/query_plan/query_planning_traversal.rs
SimonSapin: I rewrote `reorder_first_branch` with comments to be easier to understand, made it generic in order to unit-test it, and moved the option removal into it to make expectations made of the caller easier to document.

SimonSapin:
I supposed I should have commented explicitly earlier, but the changes requested have been made.

src/query_plan/query_planning_traversal.rs
sachindshinde: @SimonSapin 
This a lot easier to understand, thanks for updating comments and writing tests!

sachindshinde (approved):

dariuszkuc:
I thought we were not going to merge this as it was deleted from JS code?

SimonSapin:
I’ve filed https://github.com/apollographql/federation-next/issues/131 to track checking that the performance difference translates in Rust too. I think it’s easier to merge this and maybe remove it later, than not merge it and maybe rebase it later.

PR #122, opened 2023-12-20
feat: selection sets should use global unique ids

dariuszkuc:
resolves: https://github.com/apollographql/federation-next/issues/103

### DIFF
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -725,7 +725,7 @@ impl NormalizedFieldSelection {
 
 impl NormalizedFragmentSpreadSelection {
     /// Copies fragment spread selection and assigns it a new unique selection ID.
-    pub(crate) fn new_copy(&self) -> Self {
+    pub(crate) fn with_unique_id(&self) -> Self {
         let mut copy = self.clone();
         copy.selection_id = SelectionId::new();
         copy
@@ -804,7 +804,7 @@ impl NormalizedFragmentSpreadSelection {
 
 impl NormalizedInlineFragmentSelection {
     /// Copies inline fragment selection and assigns it a new unique selection ID.
-    pub(crate) fn new_copy(&self) -> Self {
+    pub(crate) fn with_unique_id(&self) -> Self {
         let mut copy = self.clone();
         copy.inline_fragment.selection_id = SelectionId::new();
         copy
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -14,9 +14,23 @@ use apollo_compiler::{name, Node};
 use indexmap::{IndexMap, IndexSet};
 use linked_hash_map::{Entry, LinkedHashMap};
 use std::ops::Deref;
-use std::sync::Arc;
+use std::sync::{atomic, Arc};
 
 const TYPENAME_FIELD: Name = name!("__typename");
+
+// Global storage for the counter used to uniquely identify selections
+static NEXT_ID: atomic::AtomicUsize = atomic::AtomicUsize::new(0);
+
+// opaque wrapper of the unique selection ID type
+#[derive(Clone, Debug, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


src/query_plan/operation.rs
```diff
@@ -72,23 +86,23 @@ pub(crate) enum NormalizedSelectionKey {
         // directives applied on the field
         directives: Arc<DirectiveList>,
         // unique label/counter used to distinguish fields that cannot be merged
-        label: i32,
+        label: usize,
```
goto-bus-stop, line 89: Why are these `usize` rather than `SelectionId`?

src/query_plan/operation.rs
```diff
@@ -709,6 +724,13 @@ impl NormalizedFieldSelection {
 }
 
 impl NormalizedFragmentSpreadSelection {
+    /// Copies fragment spread selection and assigns it a new unique selection ID.
+    pub(crate) fn new_copy(&self) -> Self {
```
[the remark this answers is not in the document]
sachindshinde: The idea is that sometimes you want to preserve the selection ID (e.g. when copying an element from an operation into a graph path), but it's possible other times you don't (e.g. when you want to make a genuinely new selection with the same data).

sachindshinde (changes requested):

src/query_plan/operation.rs
```diff
@@ -14,10 +14,24 @@ use apollo_compiler::{name, Node};
 use indexmap::{IndexMap, IndexSet};
 use linked_hash_map::{Entry, LinkedHashMap};
 use std::ops::Deref;
-use std::sync::Arc;
+use std::sync::{atomic, Arc};
 
 const TYPENAME_FIELD: Name = name!("__typename");
 
+// Global storage for the counter used to uniquely identify selections
+static NEXT_ID: atomic::AtomicUsize = atomic::AtomicUsize::new(0);
```
sachindshinde, line 22: I think this needs to start at `1` instead of `0`, since we use `label` of `0` below to indicate that there is no defer usage (and `fetch_add()` returns the value prior to the add).

src/query_plan/operation.rs
sachindshinde: The gist is it can be one of two possibilities: `0` to indicate that the corresponding selection doesn't use defer, or the selection ID if the selection does use defer. (We could use enums here instead, where one variant corresponds to what `0` currently means, and the other variant stores a selection ID.)

dariuszkuc: updated to `Option<SelectionId>`

src/query_plan/operation.rs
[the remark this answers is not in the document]
goto-bus-stop: Then I'd suggest naming it something like `.clone_distinct()` or `.with_unique_id()`. `new` and `copy` are both words with conventional meanings in Rust that are different from how it's used here. `.with_*` names are pretty common for methods that only change one thing about a type.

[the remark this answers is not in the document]
dariuszkuc: renamed to `with_unique_id`

sachindshinde, line 728:
```suggestion
    pub(crate) fn with_unique_id(&self) -> Self {
```

src/query_plan/operation.rs
```diff
@@ -780,6 +803,13 @@ impl NormalizedFragmentSpreadSelection {
 }
 
 impl NormalizedInlineFragmentSelection {
+    /// Copies inline fragment selection and assigns it a new unique selection ID.
+    pub(crate) fn new_copy(&self) -> Self {
```
sachindshinde, line 807:
```suggestion
    pub(crate) fn with_unique_id(&self) -> Self {
```

src/query_plan/operation.rs
dariuszkuc: my bad....

PR #123, opened 2023-12-20
chore: a few minor fixes / improvements

goto-bus-stop:
This PR is a grab-bag of a few things I ran into while getting into API schema validation. I won't be able to finish that before holiday break so posting the generic changes first.

- Change uses of `directives.iter().any()` to `directives.has()`, which is a little easier: [`DirectiveList::has`](https://docs.rs/apollo-compiler/1.0.0-beta.11/apollo_compiler/ast/struct.DirectiveList.html#method.has)
- Return an `Iterator` instead of a `Vec` from `get_types()`, the user can still decide to `collect()` it into a vec if needed. Also adds a parallel `get_directive_definitions()`.
- Return `Option<&LinksMetadata>` instead of `&Option<LinksMetadata>`-- an option containing a reference is a much more useful type than a reference to an option. You basically always want to prefer an `Option<&T>` if you can.
- c5193a39afa8ddf50e0ea60577cd3bf80f5c4d32 sets source files on supergraphs when merging subgraphs; this will let diagnostics point to source code in the future
- Remove the `Result<>` wrapper around `SpecDefinitions`. `SpecDefinitions::add` now panics if you try to add a wrong version, which is correct as it's a programmer error. And it's now much easier to use the `JOIN_VERSIONS` etc. throughout the code.

### DIFF
--- a/src/link/spec_definition.rs
+++ b/src/link/spec_definition.rs
@@ -25,7 +25,7 @@ pub(crate) trait SpecDefinition {
         schema: &FederationSchema,
         name_in_schema: &Name,
     ) -> Result<bool, FederationError> {
-        let Some(ref metadata) = schema.metadata() else {
+        let Some(metadata) = schema.metadata() else {
             return Err(SingleFederationError::Internal {
                 message: "Schema is not a core schema (add @link first)".to_owned(),
             }
@@ -42,7 +42,7 @@ pub(crate) trait SpecDefinition {
         schema: &FederationSchema,
         name_in_schema: &Name,
     ) -> Result<bool, FederationError> {
-        let Some(ref metadata) = schema.metadata() else {
+        let Some(metadata) = schema.metadata() else {
             return Err(SingleFederationError::Internal {
                 message: "Schema is not a core schema (add @link first)".to_owned(),
             }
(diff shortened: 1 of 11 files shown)


src/link/spec_definition.rs
```diff
@@ -155,23 +155,19 @@ impl<T: SpecDefinition> SpecDefinitions<T> {
...
-                message: format!(
-                    "Cannot add definition for {} to the versions of definitions for {}",
-                    definition.to_string(),
-                    self.identity
-                ),
-            }
-            .into());
-        }
+    pub(crate) fn add(&mut self, definition: T) {
+        assert_eq!(
```
dariuszkuc, line 159: since this will now panic, I'm wondering whether there would be any situation where we would want to still return partial schema even if this fails

dariuszkuc (approved):

src/link/spec_definition.rs
goto-bus-stop: from what I can tell, this particular code is all set up by federation internally, never by users. The assertion can only fail if the `definition` has a different `identity` than its type suggests, which is a bug, and should be fixed by us so we have to find out as early and loudly as possible (ideally at compile time but a crash is the second best thing 😄 )

Issue #126, opened 2024-01-04
[Plan Selection] `FetchDependencyGraph` creation and modification


lrlna:
As part of `computeBestPlanFromClosedBranches()`, we need the following `FetchDependencyGraph`-related methods to be ported to Rust:
- `QueryPlanningTraversal.newDependencyGraph()`
- `QueryPlanningTraversal.updatedDependencyGraph()`
- `FetchDependencyGraph.clone()`

Part of #77.

PR #130, opened 2024-01-05
[Plan Selection] `FetchDependencyGraph` creation and modification

SimonSapin:
Fixes https://github.com/apollographql/federation-next/issues/126


SimonSapin:
This is complete, opening a draft PR to discuss:

src/query_plan/query_planning_traversal.rs
```diff
@@ -86,3 +88,116 @@ struct BestQueryPlanInfo {
...
+                let subgraph_name = &target_node.source;
+                // The edge tail type is one of the subgraph root type, so it has to be an ObjectType.
+                //     const rootType = edge.tail.type as ObjectType;
+                let root_type = match &target_node.type_ {
+                    QueryGraphNodeType::SchemaType(position) => match position {
+                        OutputTypeDefinitionPosition::Object(object) => object.clone().into(),
+                        _ => todo!(), // Return a FederationError?
+                    },
+                    QueryGraphNodeType::FederatedRootType(_) => todo!(), // Return a FederationError?
+                };
```
SimonSapin, line 147: TypeScript `getOrCreateRootFetchGroup` expects `CompositeType` which I assume corresponds to Rust enum `CompositeTypeDefinitionPosition`.

This code obtains `QueryGraphNodeType`. How should the `FederatedRootType` case be handled?

src/query_plan/query_planning_traversal.rs
```diff
@@ -86,3 +88,116 @@ struct BestQueryPlanInfo {
...
+            let root_type = match &query_graph_node.type_ {
+                QueryGraphNodeType::SchemaType(position) => {
+                    position.clone().try_into().map_err(|err| {
+                        // TODO: add details to the error message?
+                        // format!("Should not have condition on non-selectable type {position}")
+                        err
+                    })?
+                }
+                QueryGraphNodeType::FederatedRootType(_) => todo!(), // Return a FederationError?
+            };
```
SimonSapin, line 174: Same as above

src/query_plan/query_planning_traversal.rs
```diff
@@ -86,3 +88,116 @@ struct BestQueryPlanInfo {
...
+            let group = dependency_graph.get_or_create_root_fetch_group(
```
SimonSapin, line 175: TypeScript function `computeNonRootFetchGroups` calls `getOrCreateRootFetchGroup`. Is it correct to mix "non-root" and "root" in this way?

src/query_plan/query_planning_traversal.rs
```diff
@@ -86,3 +88,116 @@ struct BestQueryPlanInfo {
...
+
+fn compute_groups_for_tree(
+    dependency_graph: &mut FetchDependencyGraph,
+    path_tree: &OpPathTree,
+    // start_group: FetchGroup,
+    // initial_group_path: GroupPath,
+    // initial_defer_context: DeferContext,
+    // initial_context: PathContext = emptyContext,
+)
+//   -> Vec<FetchGroup>
```
SimonSapin, line 200: Do you have a suggestion for Rust data structures here?

`FetchGroup` in particular is central to TypeScript `FetchDependencyGraph` but does not appear to have an equivalent in Rust. A `FetchGroup` represents a node in the graph together with its outgoing edges, right? But Rust relies on petgraph instead

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -144,3 +144,67 @@ pub(crate) struct FetchDependencyGraphPath {
...
+        parent_type: CompositeTypeDefinitionPosition,
+    ) -> &IndexSet<NodeIndex> {
+        if self.root_nodes_by_subgraph.contains_key(subgraph_name) {
+            &self.root_nodes_by_subgraph[subgraph_name]
+        } else {
+            // TODO: is there anything more to do here than create an empty IndexSet?
+            self.on_modification();
+            self.root_nodes_by_subgraph
+                .entry(subgraph_name.clone())
+                .or_default()
```
SimonSapin, line 188: Should this create a `FetchDependencyGraphNode` node? With no edges?

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -144,3 +144,67 @@ pub(crate) struct FetchDependencyGraphPath {
...
+        self.is_reduced = false;
+        self.is_optimized = false;
+    }
+
+    pub(crate) fn get_or_create_root_fetch_group(
+        &mut self,
+        subgraph_name: &NodeStr,
+        root_kind: SchemaRootDefinitionKind,
+        parent_type: CompositeTypeDefinitionPosition,
+    ) -> &IndexSet<NodeIndex> {
```
sachindshinde, line 180: This function should be returning the equivalent of a `FetchGroup`, which is a node in the fetch dependency graph (so I think a `NodeIndex` for a node that is either fetched or created by this function).

src/query_plan/fetch_dependency_graph.rs
sachindshinde: Yep, it should create a node with no edges in this case.

src/query_plan/query_planning_traversal.rs
```diff
@@ -86,3 +88,116 @@ struct BestQueryPlanInfo {
...
+        path_tree: &OpPathTree,
+    ) -> Result<(), FederationError> {
+        // TODO: is this right?
+        // In `query-planner-js/src/buildPlan.ts`, `isRootPathTree(tree)` is dependent on `tree`,
+        // but type-Tetris gives us something dependent on `self.root_kind`.
+        let is_root_path_tree = self
+            .parameters
+            .federated_query_graph
+            .root_kinds_to_nodes()?
+            .contains_key(&self.root_kind);
```
sachindshinde, line 125: This isn't quite right, since what we'd want to check is whether the `path_tree.node` is a value of the map given by `root_kinds_to_nodes()` (and not for present/absence of the root kind in the map).

Actually it's probably easier here to just check whether `path_tree.node` is a `QueryGraphNodeType::FederatedRootType`, since that only occurs for nodes in `root_kinds_to_nodes`.

src/query_plan/query_planning_traversal.rs
sachindshinde: The `FederatedRootType`-case only occurs for root nodes of the federated query graph, so you can return error here in that case (children of federated root nodes should be `SchemaType` nodes, i.e. they should belong to some subgraph schema).

src/query_plan/query_planning_traversal.rs
sachindshinde: Similarly here, you can return error if it's `FederatedRootType`, as it shouldn't occur in this case. (To be specific, `OpPathTree`s may potentially have federated root nodes as the starting node, but shouldn't have federated root nodes along any of the rest of the paths in the path tree.)

src/query_plan/query_planning_traversal.rs
sachindshinde: The gist is that the "root" vs. "non-root" in `computeRootFetchGroups()` and `computeNonRootFetchGroups()` is referring to whether the first node in the given `OpPathTree` is a federated root node or not. The "root" in `getOrCreateRootFetchGroup` means that the created fetch dependency graph node (i.e. "fetch group") is for a subgraph-entering transition and its fetch is against a root operation type of a subgraph (as opposed to something like `getOrCreateKeyFetchGroup()`, which is for a key resolution transition and its fetch is against `_entities`).

src/query_plan/query_planning_traversal.rs
sachindshinde: We've decided to rename `FetchGroup` in the TypeScript code to `FetchDependencyGraphNode` in the Rust code, since the name `FetchGroup` never really reflected that they were just nodes in a (graph-theory) graph and has led to some confusion (if you look at `fetch_dependency_graph.rs`, there's a `PORT_NOTE` comment describing the rename on `FetchDependencyGraphNode`; usually when a TypeScript function/data structure has its name changed, I try to leave a `PORT_NOTE` mentioning it).

SimonSapin:
Alright, this time I’m more confident about what’s included in the PR. But `computeGroupsForTree` is a large function that is not ported yet.

src/query_plan/query_planning_traversal.rs
```diff
@@ -86,3 +89,108 @@ struct BestQueryPlanInfo {
...
+    }
+
+    fn updated_dependency_graph(
+        &self,
+        dependency_graph: &mut FetchDependencyGraph,
+        path_tree: &OpPathTree,
+    ) -> Result<(), FederationError> {
+        let is_root_path_tree = matches!(
+            path_tree.graph.node_weight(path_tree.node)?.type_,
+            QueryGraphNodeType::SchemaType(_)
```
sachindshinde, line 121: It's inverted here, the path tree is only a root path tree if the `type_` matches `QueryGraphNodeType::FederatedRootType(_)` instead of `SchemaType`.

src/query_plan/query_planning_traversal.rs
```diff
@@ -86,3 +89,108 @@ struct BestQueryPlanInfo {
...
+                    self.root_kind,
+                    root_type,
+                );
+                compute_nodes_for_tree(
+                    dependency_graph,
+                    &child.tree,
+                    fetch_dependency_node,
+                    Default::default(),
+                    Default::default(),
+                    Conditions::TRUE,
```
sachindshinde, line 161: Instead of `Conditions::TRUE`, you'll want an `OpGraphPathContext` here with an empty vector for conditionals (could use `default()` here as well).

src/query_plan/query_planning_traversal.rs
```diff
@@ -86,3 +89,108 @@ struct BestQueryPlanInfo {
...
+                self.root_kind,
+                root_type,
+            );
+            compute_nodes_for_tree(
+                dependency_graph,
+                path_tree,
+                fetch_dependency_node,
+                Default::default(),
+                Default::default(),
+                Conditions::TRUE,
```
sachindshinde, line 191: Similar note about `OpGraphPathContext` here as above.

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -144,3 +146,162 @@ pub(crate) struct FetchDependencyGraphPath {
     path_in_node: OpPath,
     response_path: Vec<FetchDataPathElement>,
 }
+
+#[derive(Debug, Default)]
+pub(crate) struct NodePath {
```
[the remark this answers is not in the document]
SimonSapin: My thinking was that this context is provided by being part of the `fetch_dependency_graph` module, but yeah that’s lost if importing `NodePath` in some other module. I’ll change it.

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -144,3 +146,162 @@ pub(crate) struct FetchDependencyGraphPath {
...
+        subgraph_name: NodeStr,
+        subgraph_schema: ValidFederationSchema,
+        parent_type: CompositeTypeDefinitionPosition,
+        has_inputs: bool,
+        root_kind: SchemaRootDefinitionKind,
+        // merge_at: Option<Vec<ResponsePathElement>>,
+        // defer_ref: Option<NodeStr>,
+    ) -> NodeIndex {
+        self.on_modification();
+        self.graph.add_node(Arc::new(FetchDependencyGraphNode {
```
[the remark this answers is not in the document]
SimonSapin: Since that’s exactly how the `subgraph_schema` argument was obtained when calling this, I’ve removed the argument and made this method call `schema_by_source` instead.

src/query_plan/operation.rs
```diff
@@ -387,6 +388,27 @@ impl NormalizedSelection {
...
+    pub(crate) fn conditions(&self) -> Result<Conditions, FederationError> {
+        let self_conditions = Conditions::from_directives(self.directives())?;
+        if let Conditions::Boolean(false) = self_conditions {
+            // Never included, there is no point recuring
+            Ok(Conditions::Boolean(false))
+        } else {
+            match self {
+                NormalizedSelection::Field(_) => {
+                    // We want this field regardless of whether sub-selections are included
+                    Ok(self_conditions)
```
sachindshinde, line 401: The logic here doesn't match the JS codebase; when the field has a selection set and the field's conditions are non-boolean, then they should be merged into the selection set's conditions and returned (otherwise you can return `self_conditions`).

sachindshinde (changes requested):
Just one minor thing and should be good to go!

src/query_plan/fetch_dependency_graph.rs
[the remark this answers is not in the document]
sachindshinde: Makes sense!

src/query_plan/operation.rs
SimonSapin: I’ve changed to match the JS code. However the JS comment says:

> If it is a field, […] no matter what the sub-selection is, we need to get that field

Why does this only apply to boolean conditions?

sachindshinde: That's a good point! Looking at the callsites of this code, it's used to inform whether query plan nodes should be wrapped in condition nodes (i.e., whether fetches should be skipped entirely). Given that, the sub-selections of a field should indeed not affect whether the field needs to be queried.

This appears to be a bug in query planning code. I've pushed a commit to revert back to your fix, and left a `PORT_NOTE` mentioning the bug. Will merge the code in afterwards!

sachindshinde (approved):

PR #134, opened 2024-01-11
Implement `to_api_schema()`

goto-bus-stop:
Implements `to_api_schema()` and the `@inaccessible` spec. API schema generation in federation-next has 3 steps:
- Validate/remove `@inaccessible` elements (`src/link/inaccessible_spec_definition.rs`)
- Validate/remove core feature elements (imported with `@link`, `src/api_schema.rs`)
- Compatibility cleanup so the result can be compared with `graphql-js` based API schema generation (`src/compat.rs`)
  - We probably want the non-semantic directive removals to be a permanent feature, but it's currently in the `compat` module.

Tasks:
- [x] Still porting tests.
- Should use new error types -- Will do this in a separate diagnostics PR https://github.com/apollographql/federation-next/tree/renee/api-schema-diagnostics
- [x] Should remove core feature definitions (just reuse the code from `@inaccessible`)
- done-ish - Read through and refactor because I went a few different directions and there's likely some weird code left over from that. Also to make sure everything is commented appropriately 🤓


tests/api_schema.rs
```diff
@@ -0,0 +1,1728 @@
...
+#[test]
+fn removes_inaccessible_object_types() {
+    let api_schema = inaccessible_to_api_schema(
+        r#"
+      extend schema {
+        mutation: Mutation
+        subscription: Subscription
+      }
+
+      # Non-inaccessible object type
```
dariuszkuc, line 88: instead of double negation -> should we just update this comment to `accessible`?

goto-bus-stop:
Harness run:
image

src/link/inaccessible_spec_definition.rs
```diff
@@ -0,0 +1,1137 @@
...
+fn field_uses_inaccessible(field: &FieldDefinition, inaccessible_directive: &Name) -> bool {
+    field.directives.has(inaccessible_directive)
+        || field
+            .arguments
+            .iter()
+            .any(|argument| argument.directives.has(inaccessible_directive))
+}
+
+/// Check if a type definition uses the @inaccessible directive anywhere in its definition.
+fn type_uses_inaccessible(
```
dariuszkuc, line 139: since this is only used for validating core feature types should we rename it to more specific?

src/link/inaccessible_spec_definition.rs
```diff
@@ -0,0 +1,1137 @@
...
+            )+
+        }
+    }
+
+    match position {
+        TypeDefinitionPosition::Scalar(scalar_position) => {
+            let referencers = referencers
+                .scalar_types
+                .get(&scalar_position.type_name)
+                .unwrap();
```
dariuszkuc, line 735: I guess it should never be the case that this would panic but just wondering whether we should be more defensive here just in case

src/link/inaccessible_spec_definition.rs
```diff
@@ -0,0 +1,1137 @@
...
+    if has_inaccessible_field && !has_accessible_field {
+        errors.push(SingleFederationError::OnlyInaccessibleChildren {
+            message: format!("Type `{type_position}` is in the API schema but all of its members are @inaccessible."),
+        }.into());
+    }
+
+    Ok(())
+}
+
+fn validate_inaccessible_in_values(
```
dariuszkuc, line 497: maybe rename to `validate_inaccessible_in_enum_values`

dariuszkuc (changes requested):

examples/api_schema.rs
```diff
@@ -0,0 +1,25 @@
+use apollo_compiler::Schema;
+use apollo_federation::Supergraph;
+use std::process::ExitCode;
+
+fn main() -> ExitCode {
```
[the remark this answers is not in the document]
goto-bus-stop: Yeah it's handy for manual tests and as a kind of documentation, imo makes sense to have a simple example program for most individual features of crate.

src/api_schema.rs
```diff
@@ -0,0 +1,138 @@
...
+    Ok(())
+}
+
+pub fn to_api_schema(schema: FederationSchema) -> Result<Valid<Schema>, FederationError> {
+    let mut api_schema = schema;
+
+    // As we compute the API schema of a supergraph, we want to ignore explicit definitions of `@defer` and `@stream` because
+    // those correspond to the merging of potential definitions from the subgraphs, but whether the supergraph API schema
+    // supports defer or not is unrelated to whether subgraphs support it.
+    if let Some(defer) = api_schema.get_directive_definition(&name!("defer")) {
```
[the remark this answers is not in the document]
goto-bus-stop: This is how it's done in JS, I think it's specifically relevant for these because the API schema generation *doesn't know* if defer and stream should exist and relies on the router to add them if they should, while the other spec directives always exist and so the shadowing is harmless. At least that's my understanding of it

src/link/inaccessible_spec_definition.rs
goto-bus-stop: Yeah this would only happen if there is a bug, so I don't think panicking is _wrong_, but it's different from how we treat this case elsewhere, so I can change it to match the rest

src/link/inaccessible_spec_definition.rs
goto-bus-stop: it's currently only used for core feature types but the functionality isn't specific to core feature types.

src/link/inaccessible_spec_definition.rs
goto-bus-stop: I've inlined it since there's only one usage site and the arguments list is almost as long as the implementation itself :)

dariuszkuc (approved):

PR #135, opened 2024-01-11
Adjust the operation API to prevent directly setting selection keys

sachindshinde:
This PR:
- Adjusts the layout for selection data structures (field/fragment spread/inline fragment) so that they compute the key at construction-time.
- Makes key-related data for a selection immutable in those selections.
  - To do this, we create sub-modules with those structs, which prevents other code/algorithms from modifying sub-module-local struct fields.
- Makes the `NormalizedSelectionMap` API safe, in the sense that it's no longer possible to modify the key-related data of a selection once it's in the map (maintaining the invariant).
  - The previous API allowed modifying the maps's values arbitrarily (even in ways that change the computed keys), but the map's key would remain stale.

Unrelatedly, this PR also:
- Tweaks fields to not have selection IDs (the JS codebase didn't have them, after talking with Sylvain a while back I thought this was a bug, but it turns out it's because `@defer` can't be on fields).
- Adjusts keys to separate the `@defer` vs. non-`@defer` case to be separate enum variants.


src/query_graph/field_set.rs
```diff
@@ -19,36 +47,25 @@ pub(super) fn parse_field_set(
         value.as_str(),
         "field_set.graphql",
     )?;
-    merge_selection_sets(&[&field_set.selection_set])
+    Ok(field_set.into_inner().selection_set)
 }
 
 pub(super) fn merge_selection_sets(
```
dariuszkuc, line 53: since we define `NormalizedSelectionSet` in `operation.rs` -> should we move this function (and `equal_selection_sets`) over there?

src/query_plan/operation.rs
```diff
@@ -612,13 +897,6 @@ impl NormalizedSelectionSet {
 }
 
 impl NormalizedFieldSelection {
```
dariuszkuc, line 899: since you created separate modules for each selection -> should those impls move to the corresponding module space?

sachindshinde: @dariuszkuc 
Some of the `impl`s contain algorithms that shouldn't use the underlying struct fields, and should instead use the field's methods. To get that behavior, the `impl`s are outside the files. (We could alternatively have a bunch of top-level functions instead of `impl`s, but that feels a bit disorganized.)

src/query_graph/field_set.rs
sachindshinde: @dariuszkuc 
Makes sense, moved the functions over.

dariuszkuc:
Just wondering -> do we actually need those heavily nested data structures? Should we try to simplify it? i.e. `NormalizedFieldSelection  -> NormalizedField -> NormalizedFieldData` seems like it might be an overkill? 

Similarly while the regular map before could have been potentially modified, it was internal to the crate so outside manipulations were unlikely. Guess this will prevent someone from accidentally doing something odd within the lib but it does add complexity.

sachindshinde:
@dariuszkuc 
The reason for `NormalizedFieldSelection` vs `NormalizedField` is that one doesn't have selection sets and attachments, and can go in graph paths (while the other goes in operations). The reason for `NormalizedField` vs `NormalizedFieldData` is to get the immutability guarantees we want.

We could implement `Deref` on the `NormalizedField` if you'd like, that way we won't have to call `.data()` all the time.

src/query_plan/operation.rs
```diff
@@ -60,13 +62,113 @@ pub(crate) struct NormalizedSelectionSet {
...
+    /// are  merged during the normalization process. By storing a selection set as a map, we can
+    /// efficiently merge/join multiple selection sets.
+    ///
+    /// Because the key depends strictly on the value, we expose the underlying map's API in a
+    /// read-only capacity, while mutations use an API closer to `IndexSet`. We don't just use an
+    /// `IndexSet` since key computation is expensive (it involves sorting). This type is in its own
+    /// module to prevent code from accidentally mutating the underlying map outside the mutation
+    /// API.
+    ///
+    /// Note that this must be a `LinkedHashMap` so that removals don't change the order.
```
SimonSapin, line 102: We can also preserve order in `IndexMap` by using `shift_remove` instead of `swap_remove`.  I’ve filed https://github.com/apollographql/federation-next/issues/140

SimonSapin (approved):

src/query_plan/operation.rs
sachindshinde: This was easy enough, so I made the change in https://github.com/apollographql/federation-next/pull/135/commits/7daa28f06618b0b087b563cedc01ce3d9e247ba1

PR #136, opened 2024-01-15
deps: update apollo-compiler to 1.0.0-beta.12

goto-bus-stop:
Just some papercut fixes and a way to lookup schema coordinates, which is mostly intended for tests:
```rust
assert!(coord!(Type.field).lookup_field(&schema).is_ok());
```

### DIFF
--- a/src/merge.rs
+++ b/src/merge.rs
@@ -383,9 +383,8 @@ impl Merger {
                     &mut supergraph_field.make_mut().description,
                     &field.description,
                 );
-                let mut existing_args = supergraph_field.arguments.iter();
                 for arg in field.arguments.iter() {
-                    if let Some(_existing_arg) = &existing_args.find(|a| a.name == arg.name) {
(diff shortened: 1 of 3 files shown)


src/merge.rs
```diff
@@ -383,9 +383,8 @@ impl Merger {
                     &mut supergraph_field.make_mut().description,
                     &field.description,
                 );
-                let mut existing_args = supergraph_field.arguments.iter();
                 for arg in field.arguments.iter() {
-                    if let Some(_existing_arg) = &existing_args.find(|a| a.name == arg.name) {
+                    if let Some(_existing_arg) = supergraph_field.argument_by_name(&arg.name) {
```
goto-bus-stop, line 387: This was subtly wrong before: by reusing the same iterator, it keeps iterating the `existing_args` iterator where it left off. It would have the wrong result if the arguments have a different order than expected. It didn't matter because the only thing depending on this is a todo comment, but still

lrlna (approved):

Issue #128, opened 2024-01-04
[Plan Selection] `OpPathTree` creation and modification


sachindshinde:
As part of `computeBestPlanFromClosedBranches()`, we need the following `OpPathTree`-related methods to be ported to Rust:
- `PathTree.createOp()`
- `PathTree.createFromOpPaths()`
- `PathTree.merge()`

Part of #77.

PR #137, opened 2024-01-15
[Plan Selection] OpPathTree creation and modification

SimonSapin:
Fixes https://github.com/apollographql/federation-next/issues/128

### DIFF
--- a/src/query_graph/graph_path.rs
+++ b/src/query_graph/graph_path.rs
@@ -199,7 +199,8 @@ where
     TEdge: Copy + Into<Option<EdgeIndex>>,
 {
     pub(crate) fn iter(&self) -> impl Iterator<Item = GraphPathItem<'_, TTrigger, TEdge>> {
-        // FIXME: are these 3 `Vec`s expected to always have the same length?
+        debug_assert_eq!(self.edges.len(), self.edge_triggers.len());
+        debug_assert_eq!(self.edges.len(), self.edge_conditions.len());
         self.edges
             .iter()
             .copied()
--- a/src/query_graph/graph_path.rs
+++ b/src/query_graph/graph_path.rs
@@ -57,7 +57,7 @@ use std::sync::Arc;
 pub(crate) struct GraphPath<TTrigger, TEdge>
 where
     TTrigger: Eq + Hash,
-    TEdge: Into<Option<EdgeIndex>>,
+    TEdge: Copy + Into<Option<EdgeIndex>>,
 {
     /// The query graph of which this is a path.
     graph: Arc<QueryGraph>,
@@ -113,6 +113,10 @@ pub(crate) struct SubgraphEnteringEdgeInfo {
     /// The cost of resolving the conditions for this edge.
     conditions_cost: QueryPlanCost,
 }
+
+/// The item type for [`GraphPath::iter`]
+pub(crate) type GraphPathItem<'path, TTrigger, TEdge> =
+    (TEdge, &'path Arc<TTrigger>, Option<&'path Arc<OpPathTree>>);
 
 /// A `GraphPath` whose triggers are operation elements (essentially meaning that the path has been
 /// guided by a GraphQL operation).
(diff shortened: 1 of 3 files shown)


src/query_graph/path_tree.rs
```diff
@@ -31,17 +40,263 @@ pub(crate) struct PathTree<TTrigger: Eq + Hash, TEdge: Into<Option<EdgeIndex>>>
...
+                &Arc<TTrigger>,
+                Option<Arc<OpPathTree>>,
+                Vec<(
+                    /* impl Iterator<Item = GraphPathItem<…>> */ _,
+                    &'paths Arc<NormalizedSelectionSet>,
+                )>,
+            )>,
+        > = std::iter::repeat_with(Vec::new)
+            .take(edge_count + 1)
+            .collect();
```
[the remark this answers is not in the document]
SimonSapin: The first commit was a direct port. In subsequent commits I’ve refactored to use `IndexMap` in a way that hopefully makes the logic more obvious.

src/query_graph/graph_path.rs
```diff
@@ -189,6 +193,22 @@ pub(crate) struct ClosedBranch(pub(crate) Vec<Arc<ClosedPath>>);
...
+    TEdge: Copy + Into<Option<EdgeIndex>>,
+{
+    pub(crate) fn iter(&self) -> impl Iterator<Item = GraphPathItem<'_, TTrigger, TEdge>> {
+        // FIXME: are these 3 `Vec`s expected to always have the same length?
+        self.edges
+            .iter()
+            .copied()
+            .zip(&self.edge_triggers)
+            .zip(&self.edge_conditions)
+            .map(|((edge, trigger), condition)| (edge, trigger, condition))
```
SimonSapin, line 208: `zip` stops when either input iterator stops and silently truncates the rest, so it’s best with inputs of equal lengths. Is this the case here?

sachindshinde: Yeah, those three `Vec`s should always be the same length.

src/query_graph/path_tree.rs
```diff
@@ -31,17 +40,244 @@ pub(crate) struct PathTree<TTrigger: Eq + Hash, TEdge: Into<Option<EdgeIndex>>>
...
+    }
+
+    pub(crate) fn from_op_paths(
+        graph: Arc<QueryGraph>,
+        node: NodeIndex,
+        paths: &[(&OpGraphPath, &Arc<NormalizedSelectionSet>)],
+    ) -> Result<Self, FederationError> {
+        assert!(
+            !paths.is_empty(),
+            "Should compute on empty paths" // FIXME: what does this mean?
```
sachindshinde, line 79: It's like a typo, and means "Should not compute on empty paths"

src/query_graph/path_tree.rs
```diff
@@ -31,17 +40,244 @@ pub(crate) struct PathTree<TTrigger: Eq + Hash, TEdge: Into<Option<EdgeIndex>>>
...
+                        conditions: conditions.clone(),
+                        sub_paths_and_selections: vec![(graph_path_iter, selection)],
+                    });
+                }
+            }
+        }
+
+        let childs = merged
+            .into_iter()
+            .map(|(key, value)| {
```
sachindshinde, line 155: The ordering of `Vec` items here will end up being different from the JS code.

The JS code groups items with the same edge (but different triggers) together, in the order in which the triggers were encountered. The JS code also records the order in which edges were encountered for the first time, and orders these same-edge groups according to that.

sachindshinde (changes requested):

src/query_graph/path_tree.rs
SimonSapin: I’ve done that but with an intermediate struct. Please take a look.

src/query_graph/path_tree.rs
```diff
@@ -31,17 +40,249 @@ pub(crate) struct PathTree<TTrigger: Eq + Hash, TEdge: Into<Option<EdgeIndex>>>
...
+        TEdge: 'inputs,
+    {
+        // Group by and order by unique edge ID, and among those by unique trigger
+        let mut merged = IndexMap::<TEdge, ByUniqueEdge<TTrigger, /* impl Iterator */ _>>::new();
+
+        struct ByUniqueEdge<'inputs, TTriger, GraphPathIter> {
+            target_node: NodeIndex,
+            by_unique_trigger:
+                IndexMap<&'inputs Arc<TTriger>, PathTreeChildInputs<'inputs, GraphPathIter>>,
+        }
```
sachindshinde, line 116: nit: Spelling
```suggestion
        struct ByUniqueEdge<'inputs, TTrigger, GraphPathIter> {
            target_node: NodeIndex,
            by_unique_trigger:
                IndexMap<&'inputs Arc<TTrigger>, PathTreeChildInputs<'inputs, GraphPathIter>>,
        }
```

sachindshinde (approved):
LGTM, just one minor spelling nit below

PR #138, opened 2024-01-16
[operation normalization] update method to return NormalizedOperation

dariuszkuc:


dariuszkuc:
For rebasing logic I'm using `NormalizedFragment`s. Since they contain selection sets that needs to be "normalized" (but don't expand references to any other fragments) IMHO the simplest way to do it was to just add that extra boolean param `inline_fragment_spreads`. If folks have better idea let me know.

dariuszkuc:
will open separate PR that is rebased with changes from #135

PR #139, opened 2024-01-18
fix: @link compatibility with JS federation

goto-bus-stop:
Fixes two compatibility issues with our `@link` parsing found in customer schemas:
* explicit import of a root directive `@link(url: "https://example.com/myDirective", import: ["@myDirective"])` was treated as a conflict between the explicit and implicit import of `@myDirective`, changing it to only be a conflict if the explicit and implicit import came from *different* `@link`s
* linking a spec with an invalid GraphQL name, `@link(url: "https://example.com/spec-name", import: ["@directive"])`. According to the `@link` spec names must be valid GraphQL but the federation JS code does not check this and schemas exist in the wild that do this. namespacing doesn’t work with invalid names, but explicitly importing elements does, so I’m changing the Rust code to accept invalid names as well (currently with `Name::new_unchecked`, the more correct solution would be to make `name` optional, but then we have to handle this edge case everywhere… another possibility is to transform the name and make it valid somehow)

### DIFF
--- a/src/link/database.rs
+++ b/src/link/database.rs
@@ -257,8 +257,8 @@ mod tests {
 
         let meta = links_metadata(&schema)
             // TODO: error handling?
-            .unwrap_or_default()
-            .unwrap_or_default();
+            .unwrap()
+            .unwrap();
         let names_in_schema = meta
             .all_links()
             .iter()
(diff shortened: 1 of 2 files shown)


goto-bus-stop:
Given our discussion I'm leaning towards reverting the second change for now.

PR #144, opened 2024-01-22
Query Graph Traversal: Part 1

sachindshinde:
Part of #30

### DIFF
--- a/src/link/graphql_definition.rs
+++ b/src/link/graphql_definition.rs
@@ -7,6 +7,11 @@ pub(crate) struct DeferDirectiveArguments {
     if_: Option<BooleanOrVariable>,
 }
 
+/// This struct is meant for recording the original structure/intent of `@skip`/`@include`
+/// applications within the elements of a `GraphPath`. Accordingly, the order of them matters within
+/// a `Vec`, and superfluous struct instances aren't elided; `Conditions` is the more appropriate
+/// struct when trying to evaluate `@skip`/`@include` conditions (e.g. merging and short-circuiting
+/// logic).
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub(crate) struct OperationConditional {
     pub(crate) kind: OperationConditionalKind,
--- a/src/link/graphql_definition.rs
+++ b/src/link/graphql_definition.rs
@@ -9,13 +9,24 @@ pub(crate) struct DeferDirectiveArguments {
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub(crate) struct OperationConditional {
-    kind: OperationConditionalKind,
-    value: BooleanOrVariable,
+    pub(crate) kind: OperationConditionalKind,
+    pub(crate) value: BooleanOrVariable,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(
+    Debug,
+    Clone,
+    PartialEq,
+    Eq,
+    Hash,
+    strum_macros::Display,
+    strum_macros::EnumIter,
+    strum_macros::IntoStaticStr,
+)]
 pub(crate) enum OperationConditionalKind {
+    #[strum(to_string = "include")]
     Include,
+    #[strum(to_string = "skip")]
     Skip,
 }
 
--- a/src/query_graph/condition_resolver.rs
+++ b/src/query_graph/condition_resolver.rs
@@ -1,0 +1,41 @@
+use crate::query_graph::graph_path::{
+    ExcludedConditions, ExcludedDestinations, OpGraphPathContext,
+};
+use crate::query_graph::path_tree::OpPathTree;
+use crate::query_plan::QueryPlanCost;
+use petgraph::graph::EdgeIndex;
+
+/// Note that `ConditionResolver`s are guaranteed to be only called for edge with conditions.
+pub(crate) trait ConditionResolver {
+    fn resolve(
+        edge: EdgeIndex,
+        context: OpGraphPathContext,
+        excluded_destinations: ExcludedDestinations,
+        excluded_conditions: ExcludedConditions,
+    ) -> ConditionResolution;
+}
+
+pub(crate) struct ConditionResolution {
+    satisfied: bool,
+    cost: QueryPlanCost,
+    path_tree: Option<OpPathTree>,
+    // Note that this is not guaranteed to be set even if satistied === false.
+    unsatisfied_condition_reason: Option<UnsatisfiedConditionReason>,
+}
+
+pub(crate) enum UnsatisfiedConditionReason {
+    NoPostRequireKey,
+}
+
+pub(crate) struct CachingConditionResolver;
+
+impl ConditionResolver for CachingConditionResolver {
+    fn resolve(
+        _edge: EdgeIndex,
+        _context: OpGraphPathContext,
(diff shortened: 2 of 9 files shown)


src/query_graph/graph_path.rs
```diff
@@ -238,4 +481,770 @@ impl OpGraphPath {
...
+        // large `Vec`s and be a bottleneck, and a more iterative version that pre-allocates `Vec`s
+        // is quite a bit faster.
+        if options_for_each_path.is_empty() {
+            return vec![];
+        }
+
+        // Track, for each path, which option index we're at.
+        let mut option_indexes = std::iter::repeat_with(|| 0)
+            .take(options_for_each_path.len())
+            .collect::<Vec<usize>>();
```
SimonSapin, line 857:
```suggestion
        let mut option_indexes = vec![0; options_for_each_path.len()];
```

This relies on `usize: Copy`. The repeat_with + take + collect formulation is useful for `!Copy` values.

src/query_graph/graph_path.rs
```diff
@@ -238,4 +481,770 @@ impl OpGraphPath {
...
+
+        // Pre-allocate `Vec` for the result.
+        let num_options = options_for_each_path
+            .iter()
+            .fold(1, |acc, options| acc * options.len());
```
SimonSapin, line 862:
```suggestion
        let num_options = options_for_each_path
            .iter()
            .map(|options| options.len())
            .product();
```

src/query_graph/graph_path.rs
```diff
@@ -238,4 +481,770 @@ impl OpGraphPath {
...
+        let mut product = Vec::with_capacity(num_options);
+
+        // Compute the cartesian product.
+        for _ in 0..num_options {
+            let num_simultaneous_paths = options_for_each_path
+                .iter()
+                .zip(option_indexes.iter())
+                .fold(0, |acc, (options, option_index)| {
+                    acc + options[*option_index].0.len()
+                });
```
SimonSapin, line 872: `zip` accepts `IntoIterator`, not just `Iterator`:

```suggestion
                .zip(&option_indexes)
                .map(|(options, option_index)| {
                    options[*option_index].0.len()
                })
                .sum();
```

src/query_graph/graph_path.rs
```diff
@@ -238,4 +481,770 @@ impl OpGraphPath {
...
+                self.0[0].compare_single_path_options_complexity_out_of_context(&other.0[0])
+            }
+            (true, false) => {
+                self.0[0].compare_single_vs_multi_path_options_complexity_out_of_context(other)
+            }
+            (false, true) => Ok(other.0[0]
+                .compare_single_vs_multi_path_options_complexity_out_of_context(self)?
+                .reverse()),
+            (false, false) => Ok(Ordering::Equal),
+        }
```
SimonSapin, line 924: Slice patterns can extract the item based on the length of the slice value:
```suggestion
        match (self.0.as_slice(), other.0.as_slice()) {
            ([a], [b]) => {
                a.compare_single_path_options_complexity_out_of_context(b)
            }
            ([a], _) => {
                a.compare_single_vs_multi_path_options_complexity_out_of_context(other)
            }
            (_, [b]) => Ok(b
                .compare_single_vs_multi_path_options_complexity_out_of_context(self)?
                .reverse()),
            _ => Ok(Ordering::Equal),
        }
```

src/query_graph/mod.rs
```diff
@@ -382,4 +386,126 @@ impl QueryGraph {
...
+            };
+            if field.data().field_position == *field_definition_position {
+                candidates.push(edge_ref.id())
+            }
+        }
+        if candidates.len() > 1 {
+            Err(SingleFederationError::Internal {
+                message: format!("Node has multiple edges matching {}", field),
+            }
+            .into())
```
SimonSapin, line 413: Is it possible to end up with many candidates? Instead of finding them all could the loop end early as soon as finding the second one? (And keep an `Option` local variable instead of a `Vec`.)

SimonSapin (approved):
The suggestions are all minor, feel free to take or ignore as you like

src/query_graph/condition_resolver.rs
```diff
@@ -0,0 +1,41 @@
...
+    satisfied: bool,
+    cost: QueryPlanCost,
+    path_tree: Option<OpPathTree>,
+    // Note that this is not guaranteed to be set even if satistied === false.
+    unsatisfied_condition_reason: Option<UnsatisfiedConditionReason>,
+}
+
+pub(crate) enum UnsatisfiedConditionReason {
+    NoPostRequireKey,
+}
```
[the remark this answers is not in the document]
sachindshinde: This is a carryover from the corresponding enum in the JS codebase. My understanding is that we should be adding more variants to improve diagnostics/logging, but it hasn't been done yet.

src/link/graphql_definition.rs
```diff
@@ -9,13 +9,24 @@ pub(crate) struct DeferDirectiveArguments {
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub(crate) struct OperationConditional {
-    kind: OperationConditionalKind,
-    value: BooleanOrVariable,
+    pub(crate) kind: OperationConditionalKind,
+    pub(crate) value: BooleanOrVariable,
 }
```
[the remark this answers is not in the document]
sachindshinde: The JS codebase has both. My understanding is they serve different purposes; `OperationConditional` is meant to preserve the overall structure/intent from the user (e.g. we use a `Vec<OperationConditional>` where order matters, and don't short-circuit), while `Conditions` is more for actually evaluating/short-circuiting the conditions.

src/query_graph/graph_path.rs
sachindshinde: Ah nice, didn't know about that syntax!

src/query_graph/mod.rs
sachindshinde: The JS codebase was doing the check here. It shouldn't be possible to have multiple candidates (the progressive override work we'll have to eventually port will change the story a bit, but ultimately it's still only one candidate expected). I'm down with assuming validity here, and just returning early instead of checking the rest of the `Vec` for another match.

src/query_graph/graph_path.rs
```diff
@@ -238,4 +481,770 @@ impl OpGraphPath {
...
+        let mut runtime_types = Arc::new(match &head_weight.type_ {
+            QueryGraphNodeType::SchemaType(head_type_pos) => {
+                let head_type_pos: CompositeTypeDefinitionPosition =
+                    head_type_pos.clone().try_into()?;
+                self.graph
+                    .schema_by_source(&head_weight.source)?
+                    .possible_runtime_types(head_type_pos)?
+            }
+            QueryGraphNodeType::FederatedRootType(_) => IndexSet::new(),
+        });
```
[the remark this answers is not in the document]
sachindshinde: Makes sense, extracted it as `GraphPath::head_possible_runtime_types()`.

src/link/graphql_definition.rs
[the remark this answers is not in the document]
dariuszkuc: can we add comments on struct to indicate this?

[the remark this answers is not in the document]
sachindshinde: Yep, can add comments there.

Issue #91, opened 2023-11-17
[operation processing] rebase fragments on subgraph fetches


dariuszkuc:
**This ticket has been moved to https://apollographql.atlassian.net/browse/FED-32.**

---

lrlna:
There are two components to this:
- rebasing operations (important for qp)
- reusing fragments (optimisation)

PR #157, opened 2024-01-23
[operation normalization] return NormalizedOperation and normalize Fragment

dariuszkuc:
`normalize_operation` now returns `NormalizedOperation`.

By default, when normalizing selection sets, we auto expand and inline fragment spreads. In order to be able to rebase fragments we need to normalize their selection sets without inlining other fragments. Updated method to accept explicit `FragmentSpreadNormalizationOption` enum option to indicate whether to inline spreads or preserve them (it is the same logic except for handling fragment spreads).

### DIFF
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -494,7 +494,7 @@ impl NormalizedFragment {
                 &fragment.selection_set,
                 &IndexMap::new(),
                 schema,
-                &FragmentSpreadNormalizationOption::PreserveFragmentSpread,
+                FragmentSpreadNormalizationOption::PreserveFragmentSpread,
             )?,
         })
     }
@@ -748,6 +748,7 @@ pub(crate) mod normalized_inline_fragment_selection {
 }
 
 /// Available fragment spread normalization options
+#[derive(Copy, Clone)]
 pub(crate) enum FragmentSpreadNormalizationOption {
     InlineFragmentSpread,
     PreserveFragmentSpread,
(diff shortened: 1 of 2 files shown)


src/query_plan/operation.rs
```diff
@@ -416,6 +418,28 @@ pub(crate) struct NormalizedFragment {
...
+                .get_type(fragment.type_condition().clone())?
+                .try_into()?,
+            directives: Arc::new(fragment.directives.clone()),
+            selection_set: NormalizedSelectionSet::normalize_and_expand_fragments(
+                &fragment.selection_set,
+                &IndexMap::new(),
+                schema,
+                false,
+            )?,
+        })
```
SimonSapin, line 439: A downside of boolean arguments is that when a caller passes `true` or `false` there’s no indication of what that mean. Consider either:

* Replace `bool` with a dedicated two-variant enum
* Or introduce a named variable at such call sites. For example:

```suggestion
        let inline_fragment_spreads = false;
        Ok(Self {
            schema: schema.clone(),
            name: fragment.name.clone(),
            type_condition_position: schema
                .get_type(fragment.type_condition().clone())?
                .try_into()?,
            directives: Arc::new(fragment.directives.clone()),
            selection_set: NormalizedSelectionSet::normalize_and_expand_fragments(
                &fragment.selection_set,
                &IndexMap::new(),
                schema,
                inline_fragment_spreads,
            )?,
        })
```

dariuszkuc: indeed, i like the explicit enum variant -> will update it

lrlna:
@dariuszkuc this was partially for #91, right?

PR #160, opened 2024-01-30
feat: add options to insert `@defer`, `@stream` in API schema

goto-bus-stop:
Whether `@defer` or `@stream` should be exposed depends on the capabilities and configuration of the supergraph router. This option adds the definitions to the API schema if requested.


dariuszkuc:
Just wondering -> shouldn't we just always add it in? I believe in the past we had execution paths for including/excluding them based on whether it was Router (which supports it) and Gateway (which doesn't)

goto-bus-stop:
it might be the right answer eventually, at the moment it's still possible to disable defer entirely in router: https://www.apollographql.com/docs/router/executing-operations/defer-support/#disabling-defer

PR #161, opened 2024-02-01
Query Graph Traversal: Part 2

### LINKED ISSUE #155: [Graph Traversal] advance graph path of operation element
split off #30

sachindshinde:
Fixes #155

### DIFF
--- a/src/link/argument.rs
+++ b/src/link/argument.rs
@@ -1,4 +1,5 @@
 use crate::error::{FederationError, SingleFederationError};
+use crate::link::graphql_definition::BooleanOrVariable;
 use apollo_compiler::ast::Value;
 use apollo_compiler::schema::{Directive, Name};
 use apollo_compiler::{Node, NodeStr};
@@ -140,4 +141,22 @@ pub(crate) fn directive_required_boolean_argument(
         }
         .into()
     })
+}
+
+pub(crate) fn directive_optional_variable_boolean_argument(
+    application: &Node<Directive>,
+    name: &Name,
+) -> Result<Option<BooleanOrVariable>, FederationError> {
+    match application.argument_by_name(name) {
+        Some(value) => match value.deref() {
+            Value::Variable(name) => Ok(Some(BooleanOrVariable::Variable(name.clone()))),
+            Value::Boolean(value) => Ok(Some(BooleanOrVariable::Boolean(*value))),
+            Value::Null => Ok(None),
+            _ => Err(FederationError::internal(format!(
+                "Argument \"{}\" of directive \"@{}\" must be a boolean.",
+                name, application.name
+            ))),
+        },
+        None => Ok(None),
+    }
 }
--- a/src/query_graph/condition_resolver.rs
+++ b/src/query_graph/condition_resolver.rs
@@ -18,14 +18,15 @@ pub(crate) trait ConditionResolver {
     ) -> Result<ConditionResolution, FederationError>;
 }
 
-// TODO: This could probably be refactored into an enum.
 #[derive(Debug, Clone)]
-pub(crate) struct ConditionResolution {
-    pub(crate) satisfied: bool,
-    pub(crate) cost: QueryPlanCost,
-    pub(crate) path_tree: Option<Arc<OpPathTree>>,
-    // Note that this is not guaranteed to be set even if satisfied is false.
-    pub(crate) unsatisfied_condition_reason: Option<UnsatisfiedConditionReason>,
+pub(crate) enum ConditionResolution {
+    Satisfied {
+        cost: QueryPlanCost,
+        path_tree: Option<Arc<OpPathTree>>,
+    },
+    Unsatisfied {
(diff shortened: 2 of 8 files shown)


src/query_graph/mod.rs
```diff
@@ -393,30 +394,92 @@ impl QueryGraph {
...
     }
 
+    pub(crate) fn edge_for_field(
+        &self,
+        node: NodeIndex,
+        field: &NormalizedField,
+    ) -> Option<EdgeIndex> {
+        // PORT_NOTE: The JS codebase found all matching candidates and asserted there was only one.
+        // We consider it unlikely for there to be more than one candidate given all the code paths
+        // that create edges, so we've removed the assert here and return on the first match.
```
SimonSapin, line 432: We could still assert (or debug assert) there isn’t a second candidate without collecting all of them:

```rust
let mut iter = /* … */;
if let Some(first) = iter.next() {
    debug_assert!(iter.next().is_none(), "unexpectedly found multiple candidates")
    // …
}

File: src/query_graph/mod.rs [hunk 2]
```diff
@@ -504,4 +567,38 @@ impl QueryGraph {
...
+        };
+        Ok(*is_part_of_provides)
+    }
+
+    pub(crate) fn has_an_implementation_with_provides(
+        &self,
+        _source: &NodeStr,
+        _interface_field_definition_position: InterfaceFieldDefinitionPosition,
+    ) -> Result<bool, FederationError> {
+        todo!()
```
SimonSapin, line 602: we’ll definitely need to grep the entire code base for `FIXME`/`TODO`/`todo!` at some point

src/query_plan/operation.rs
```diff
@@ -38,7 +38,7 @@ static NEXT_ID: atomic::AtomicUsize = atomic::AtomicUsize::new(1);
 pub(crate) struct SelectionId(usize);
 
 impl SelectionId {
-    fn new() -> Self {
+    pub(crate) fn new() -> Self {
         // atomically increment global counter
         Self(NEXT_ID.fetch_add(1, atomic::Ordering::AcqRel))
```
SimonSapin, line 43: Same as above for `OverrideId` re serialization

src/query_plan/operation.rs
```diff
@@ -704,6 +707,18 @@ pub(crate) mod normalized_inline_fragment_selection {
         pub(crate) selection_id: SelectionId,
     }
 
+    impl NormalizedInlineFragmentData {
+        pub(crate) fn defer_directive_arguments(
+            &self,
+        ) -> Result<Option<DeferDirectiveArguments>, FederationError> {
+            if let Some(directive) = self.directives.get(&name!("defer")) {
```
SimonSapin, line 714:
```suggestion
            if let Some(directive) = self.directives.get("defer") {
```

I think this should be equivalent

SimonSapin (approved):

src/link/argument.rs
```diff
@@ -141,3 +142,21 @@ pub(crate) fn directive_required_boolean_argument(
         .into()
     })
 }
+
+pub(crate) fn directive_optional_variable_boolean_argument(
```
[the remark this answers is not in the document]
sachindshinde: Yeah, having something cleaner/simpler here would be nice (ideally with some way to handle custom scalars).

src/query_graph/condition_resolver.rs
```diff
@@ -1,41 +1,68 @@
...
+            cost: 0,
+            path_tree: None,
+            unsatisfied_condition_reason: None,
+        }
+    }
+
+    pub(crate) fn unsatisfied_conditions() -> Self {
+        Self {
+            satisfied: false,
+            cost: -1,
```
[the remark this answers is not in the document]
sachindshinde: It should never be used directly if it's `-1` I think. I ended up refactoring this into an enum instead, so `cost` vanishes in the unsatisfied case.

src/query_graph/graph_path.rs
```diff
@@ -127,6 +155,21 @@ pub(crate) struct SubgraphEnteringEdgeInfo {
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub(crate) struct OverrideId(usize);
+
+/// Global storage for the counter used to allocate `OverrideId`s.
+static NEXT_OVERRIDE_ID: atomic::AtomicUsize = atomic::AtomicUsize::new(1);
+
+impl OverrideId {
+    fn new() -> Self {
+        // atomically increment global counter
+        Self(NEXT_OVERRIDE_ID.fetch_add(1, atomic::Ordering::AcqRel))
```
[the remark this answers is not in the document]
sachindshinde: Nah, we shouldn't be serializing/deserializing these AFAIK. Will add a warning in a comment.

src/query_graph/mod.rs
sachindshinde: That's a nice pattern, will add the `debug_assert!()`!

Issue #152, opened 2024-01-22
[Plan Selection] compute fetch groups for op path tree


lrlna:
Split this off from the update function in #126 to make it easier to port

Issue #153, opened 2024-01-22
[Plan Selection] handle requires for fetch dependency graph


lrlna:
**This ticket has been moved to https://apollographql.atlassian.net/browse/FED-25.**

---

also a split off from #126

SimonSapin:
https://github.com/apollographql/federation-next/pull/172 adds an `fn handle_requires(…)` stub in `src/query_plan/fetch_dependency_graph.rs`

PR #172, opened 2024-02-07
Port computeGroupsForTree from buildPlan.ts

SimonSapin:
Fixes https://github.com/apollographql/federation-next/issues/152

### DIFF
--- a/src/query_plan/fetch_dependency_graph.rs
+++ b/src/query_plan/fetch_dependency_graph.rs
@@ -1243,5 +1243,7 @@ fn handle_requires(
     _defer_context: &DeferContext,
     _created_nodes: &mut IndexSet<NodeIndex>,
 ) -> Result<(NodeIndex, FetchDependencyGraphNodePath), FederationError> {
+    // PORT_NOTE: instead of returing IDs of created nodes they should be inserted directly
+    // in the `created_nodes` set passed by mutable reference.
     todo!() // Port `handleRequires`
 }
--- a/src/query_plan/fetch_dependency_graph.rs
+++ b/src/query_plan/fetch_dependency_graph.rs
@@ -773,24 +773,33 @@ pub(crate) fn compute_nodes_for_tree(
                         )));
                     };
                     let edge = stack_item.tree.graph.edge_weight(edge_id)?;
-                    if let QueryGraphEdgeTransition::KeyResolution = &edge.transition {
-                        stack.push(compute_nodes_for_key_resolution(
-                            dependency_graph,
-                            &stack_item,
-                            child,
-                            edge_id,
-                            new_context,
-                            &mut created_nodes,
-                        )?);
-                    } else {
-                        stack.push(compute_nodes_for_graph_path_context(
-                            dependency_graph,
-                            &stack_item,
-                            child,
-                            edge_id,
-                            edge,
-                            new_context,
-                        )?);
+                    match edge.transition {
+                        QueryGraphEdgeTransition::KeyResolution => {
+                            stack.push(compute_nodes_for_key_resolution(
+                                dependency_graph,
+                                &stack_item,
+                                child,
+                                edge_id,
+                                new_context,
+                                &mut created_nodes,
+                            )?);
+                        }
+                        QueryGraphEdgeTransition::RootTypeResolution { root_kind } => {
+                            stack.push(compute_nodes_for_root_type_resolution(
+                                dependency_graph,
+                                &stack_item,
+                                child,
+                                edge_id,
+                                edge,
+                                root_kind,
+                                new_context,
+                            )?);
+                        }
+                        _ => {
+                            return Err(FederationError::internal(format!(
+                                "Unexpected non-collecting edge {edge}"
+                            )))
+                        }
                     }
                 }
                 OpGraphPathTrigger::OpPathElement(operation) => {
@@ -931,19 +940,15 @@ fn compute_nodes_for_key_resolution<'a>(
     })
 }
 
-fn compute_nodes_for_graph_path_context<'a>(
+fn compute_nodes_for_root_type_resolution<'a>(
     dependency_graph: &mut FetchDependencyGraph,
     stack_item: &ComputeNodesStackItem<'_>,
     child: &'a Arc<PathTreeChild<OpGraphPathTrigger, Option<EdgeIndex>>>,
     edge_id: EdgeIndex,
     edge: &crate::query_graph::QueryGraphEdge,
+    root_kind: SchemaRootDefinitionKind,
     new_context: &'a OpGraphPathContext,
 ) -> Result<ComputeNodesStackItem<'a>, FederationError> {
-    let QueryGraphEdgeTransition::RootTypeResolution { root_kind } = edge.transition else {
-        return Err(FederationError::internal(format!(
-            "Unexpected non-collecting edge {edge}"
-        )));
-    };
     if child.conditions.is_some() {
         return Err(FederationError::internal(format!(
             "Root type resolution edge {edge} should not have conditions"
--- a/src/query_plan/fetch_dependency_graph.rs
+++ b/src/query_plan/fetch_dependency_graph.rs
@@ -1,16 +1,25 @@
 use crate::error::FederationError;
-use crate::query_graph::graph_path::{OpGraphPathContext, OpPath};
-use crate::query_graph::path_tree::OpPathTree;
-use crate::query_graph::QueryGraph;
+use crate::query_graph::graph_path::{
+    OpGraphPathContext, OpGraphPathTrigger, OpPath, OpPathElement,
+};
+use crate::query_graph::path_tree::{OpPathTree, PathTreeChild};
+use crate::query_graph::{QueryGraph, QueryGraphEdgeTransition};
 use crate::query_plan::conditions::Conditions;
-use crate::query_plan::operation::NormalizedSelectionSet;
-use crate::query_plan::{FetchDataPathElement, QueryPathElement};
+use crate::query_plan::operation::normalized_field_selection::{
+    NormalizedField, NormalizedFieldData,
+};
+use crate::query_plan::operation::{NormalizedSelection, NormalizedSelectionSet, TYPENAME_FIELD};
+use crate::query_plan::FetchDataPathElement;
 use crate::query_plan::{FetchDataRewrite, QueryPlanCost};
-use crate::schema::position::{CompositeTypeDefinitionPosition, SchemaRootDefinitionKind};
+use crate::schema::position::{
+    CompositeTypeDefinitionPosition, ObjectTypeDefinitionPosition, SchemaRootDefinitionKind,
+};
 use crate::schema::ValidFederationSchema;
+use apollo_compiler::schema::{self, Name};
 use apollo_compiler::NodeStr;
 use indexmap::{IndexMap, IndexSet};
-use petgraph::stable_graph::{NodeIndex, StableDiGraph};
+use petgraph::stable_graph::{EdgeIndex, NodeIndex, StableDiGraph};
+use petgraph::visit::EdgeRef;
 use std::sync::Arc;
 
 /// Represents a subgraph fetch of a query plan.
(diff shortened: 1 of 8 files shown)


src/query_plan/fetch_dependency_graph.rs
```diff
@@ -292,18 +634,568 @@ impl DeferTracking {
...
+        )
+    }
+    let input_type = dependency_graph.type_for_fetch_inputs(source_type.type_name())?;
+    let mut input_selections = NormalizedSelectionSet::empty(
+        // TODO: should this be a subgraph schema?
+        dependency_graph.supergraph_schema.clone(),
+        input_type.clone(),
+    );
+    let Some(edge_conditions) = &edge.conditions else {
+        // TODO: Can this ever be `None`?
```
dariuszkuc, line 837: I don't think so (also in general in JS code whenever there was a chance for error it would use some validation/explicit throws for nice error messages -> non-null assert here would result in pretty bad stack trace)

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -292,18 +634,568 @@ impl DeferTracking {
...
+            dependency_graph,
+            edge_id,
+            conditions,
+            (stack_item.node_id, &stack_item.node_path),
+            stack_item.context,
+            &updated.defer_context,
+            created_nodes,
+        )?;
+        updated.node_id = required_node_id;
+        updated.node_path = require_path;
```
dariuszkuc, line 1093: do we need to also extend `created_nodes` with `requires` nodes? https://github.com/apollographql/federation/blob/main/query-planner-js/src/buildPlan.ts#L4109

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -292,18 +634,568 @@ impl DeferTracking {
...
+                updated.node_id,
+            )?;
+            updated_node.must_preserve_selection_set = true
+        }
+    }
+    let edge = child.tree.graph.edge_weight(edge_id)?;
+    if let QueryGraphEdgeTransition::InterfaceObjectFakeDownCast { .. } = &edge.transition {
+        // We shouldn't add the operation "as is" as it's a down-cast but we're "faking it".
+        // However, if the operation has directives, we should preserve that.
+        let OpPathElement::InlineFragment(inline) = updated_operation else {
```
dariuszkuc, line 1141: I think JS `FragmentElement` represents both inline fragment and named spreads -> their selections actually reference same `FragmentElement` (https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L3122)

dariuszkuc (changes requested):
There are few unimplemented methods - guessing those would be ported over in subsequent PR(s)?

I am unsure whether we should be using supergraph or subgraph schemas when creating those new selection sets/field data.

Other than it is a pretty clean port of the JS stuff over.

src/query_plan/operation.rs
```diff
@@ -1247,6 +1261,42 @@ impl NormalizedSelectionSet {
...
+    /// but when some are provided, uncesssary fragments will be automaticaly removed
+    /// at the junction between the path and those final selections.
+    /// For instance, suppose that we have:
+    ///  - a `path` argument that is `a::b::c`,
+    ///    where the type of the last field `c` is some object type `C`.
+    ///  - a `selections` argument that is `{ ... on C { d } }`.
+    /// Then the resulting built selection set will be: `{ a { b { c { d } } }`,
+    /// and in particular the `... on C` fragment will be eliminated since it is unecesasry
+    /// (since again, `c` is of type `C`).
+    pub(crate) fn add_at_path(
```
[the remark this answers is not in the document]
SimonSapin: As I understand it the point of that module is to keep `IndexMap<NormalizedSelectionKey, NormalizedSelection>` private and expose and API like a set of `NormalizedSelection`s. Code inside the module is responsible for keeping `NormalizedSelectionKey`s consistent with their values. Code outside of it doesn’t see keys at all so we don’t need to worry about getting them wrong. It can rely on lower-level methods like `insert`, which are inside of the module.

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -292,18 +634,568 @@ impl DeferTracking {
...
+        node_id: start_node_id,
+        node_path: initial_node_path,
+        context: initial_conditions,
+        defer_context: initial_defer_context,
+    }];
+    let mut created_nodes = IndexSet::new();
+    while let Some(stack_item) = stack.pop() {
+        let node =
+            FetchDependencyGraph::node_weight_mut(&mut dependency_graph.graph, stack_item.node_id)?;
+        for selection_set in &tree.local_selection_sets {
```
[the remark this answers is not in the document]
SimonSapin: Indeed this is a bug! Initially I didn’t have a `stack_item` variable but deconstructed it like JS does: `const { tree, group, path, context, deferContext } = stack.pop()!;`, shadowing the previous `tree` binding. I added `stack_item` when splitting this into more functions, because Clippy rightfully complained about taking 12 separate parameters.

I’ll rename the parameter of `compute_nodes_for_tree` to `initial_tree` to avoid this confusion.

src/query_plan/fetch_dependency_graph.rs
SimonSapin: In JS `createdGroups: FetchGroup[]` is part of the return value of `handleRequires()`. In Rust instead, `handle_requires` is still a TODO in this PR but it takes a `&mut IndexSet<NodeIndex>` parameter so it can insert directly instead of returning and intermediate set or vec.

src/query_plan/fetch_dependency_graph.rs
SimonSapin: Here `update_operation` has type:

```rust
pub(crate) enum OpPathElement {
    Field(NormalizedField),
    InlineFragment(NormalizedInlineFragment),
}
```

There is no variant for named fragments, so presumably they’ve been inlined when or before creating an `OpPathElement`.

src/query_plan/fetch_dependency_graph.rs
dariuszkuc: since it could be easily missed in subsequent PR -> can you add some comment/todo so we don't forget?

SimonSapin: I’ve added a `PORT_NOTE` comment

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -347,18 +686,564 @@ impl DeferTracking {
...
+                path_in_parent: path.as_ref(),
+            },
+        )
+    }
+    let input_type = dependency_graph.type_for_fetch_inputs(source_type.type_name())?;
+    let mut input_selections = NormalizedSelectionSet::empty(
+        // TODO: should this be a subgraph schema?
+        dependency_graph.supergraph_schema.clone(),
+        input_type.clone(),
+    );
```
SimonSapin, line 883: Unsure about this. Here we create a new selection set, which in Rust contains a reference to the schema. As far as I can tell JS has no such reference. @sachindshinde, do you know which schema this should be? Does it matter?

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -321,6 +640,18 @@ impl FetchSelectionSet {
...
+
+    fn add_at_path(
+        &mut self,
+        path_in_node: &OpPath,
+        selection_set: Option<&Arc<NormalizedSelectionSet>>,
+    ) -> Result<(), FederationError> {
+        Arc::make_mut(&mut self.selection_set).add_at_path(path_in_node, selection_set);
+        // TODO: when calling this multiple times, maybe only re-compute conditions at the end?
+        // Or make it lazily-initialized and computed on demand?
+        self.conditions = self.selection_set.conditions()?;
```
SimonSapin, line 652: Here we re-compute conditions for the entire selection set for every call to `FetchSelectionSet::add_at_path`, which could have O(n²) cost. Should conditions be lazily initialized (with `add_to_path` only emptying the cache) to avoid this?

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -330,14 +661,22 @@ impl FetchInputs {
...
         }
     }
+
+    fn add(&self, selection: &NormalizedSelectionSet) {
+        assert!(
+            selection.schema.ptr_eq(&self.supergraph_schema),
+            "Inputs selections must be based on the supergraph schema"
+        );
+        todo!()
+    }
```
SimonSapin, line 671: `FetchInputs::add` needs to be ported from JS `GroupInputs.add` in a follow-up PR

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -347,18 +686,564 @@ impl DeferTracking {
...
+}
+
+fn extract_defer_from_operation(
+    _dependency_graph: &mut FetchDependencyGraph,
+    _operation: &OpPathElement,
+    _defer_context: &DeferContext,
+    _node_path: &FetchDependencyGraphNodePath,
+) -> (Option<OpPathElement>, DeferContext) {
+    todo!() // Port `extractDeferFromOperation`
+}
```
SimonSapin, line 1235: Also needs follow-up PR(s)

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -347,18 +686,564 @@ impl DeferTracking {
...
+    _requires_conditions: &OpPathTree,
+    (_node_id, _node_path): (NodeIndex, &FetchDependencyGraphNodePath),
+    _context: &OpGraphPathContext,
+    _defer_context: &DeferContext,
+    _created_nodes: &mut IndexSet<NodeIndex>,
+) -> Result<(NodeIndex, FetchDependencyGraphNodePath), FederationError> {
+    // PORT_NOTE: instead of returing IDs of created nodes they should be inserted directly
+    // in the `created_nodes` set passed by mutable reference.
+    todo!() // Port `handleRequires`
 }
```
SimonSapin, line 1249: Porting `handle_requires` is tracked at https://github.com/apollographql/federation-next/issues/153

src/query_plan/operation.rs
```diff
@@ -1368,6 +1382,42 @@ impl NormalizedSelectionSet {
...
+        &mut self,
+        path: &OpPath,
+        selection_set: Option<&Arc<NormalizedSelectionSet>>,
+    ) {
+        Arc::make_mut(&mut self.selections).add_at_path(path, selection_set)
+    }
+
+    pub(crate) fn add(&mut self, _selections: &NormalizedSelectionSet) {
+        todo!()
+    }
```
SimonSapin, line 1396: Needs follow-up

src/query_plan/operation.rs
```diff
@@ -1368,6 +1382,42 @@ impl NormalizedSelectionSet {
...
+    /// and in particular the `... on C` fragment will be eliminated since it is unecesasry
+    /// (since again, `c` is of type `C`).
+    pub(crate) fn add_at_path(
+        &mut self,
+        _path: &OpPath,
+        _selection_set: Option<&Arc<NormalizedSelectionSet>>,
+    ) {
+        // TODO: port a `SelectionSetUpdates` data structure or mutate directly?
+        todo!()
+    }
```
SimonSapin, line 1420: .

src/query_plan/fetch_dependency_graph.rs
sachindshinde: @SimonSapin With the JS codebase, the schema elements themselves internally stored references to their schemas. It's those element's schemas that are relevant here.

The `typeForFetchInputs()` function in JS pulls its element from the supergraph schema, so the supergraph schema is whats used here.

There's also [a comment](https://github.com/apollographql/federation/blob/%40apollo/query-planner%402.6.3/query-planner-js/src/buildPlan.ts#L3932-L3934) in the JS code that elaborates on this; it would be nice to copy that over here.

SimonSapin: Alright, done. I’m not sure why I initially omitted them but some other comments were missing so I’ve ported them too.

dariuszkuc (approved):

dariuszkuc:
Can we double check we created issues for all todos for follow up PR?

PR #175, opened 2024-02-13
Query Planner constructor and some strictness improvements

### LINKED ISSUE #156: [QP Construction] finish porting the constructor
**This ticket has been moved to https://apollographql.atlassian.net/browse/FED-22.**

---

pre computed data structures after building the api query graph

goto-bus-stop:
This implements the constructor `QueryPlanner::new()`, which builds query graphs and does some minor precomputation, and:
- fixes a few issues in subgraph extraction, were easy to miss since that code was not used before. also removed an old module that is obsolete.
- introduce some use of `ValidFederationSchema` vs. `FederationSchema` in the API schema implementation, so there is less potential for misuse (this requires more cloning than i'd like--i'm not convinced that ValidFederationSchema being an Arc internally is the best thing to do, but understand the usefulness of its `==` implementation!)
- move `@inaccessible` related functions onto the spec definition struct: this means you have to look up the inaccessible spec before using it, manually or from the schema, so there's less potential for misuse

Closes FED-22
Closes FED-82


goto-bus-stop:
Most of the existing tests on the JS side are checking behaviours in Fed 1 supergraphs. I don't think I can port those over yet as Fed 1 supergraphs are out of scope. I will add some notes/questions to this PR about the differences between JS and Rust data structures

src/query_plan/query_planner.rs
```diff
@@ -106,26 +120,305 @@ impl Default for QueryPlannerDebugConfig {
...
 }
 
 pub struct QueryPlanner {
-    config: Arc<QueryPlannerConfig>,
+    // TODO(@goto-bus-stop) not convinced that these arcs are all necessary
+    config: QueryPlannerConfig,
     federated_query_graph: Arc<QueryGraph>,
     supergraph_schema: ValidFederationSchema,
     api_schema: ValidFederationSchema,
     subgraph_federation_spec_definitions: Arc<IndexMap<NodeStr, &'static FederationSpecDefinition>>,
```
goto-bus-stop, line 135: I didn't find this in the JS implementation. I do understand what it needs to do because of the type. But how is it going to be used?

src/lib.rs
```diff
@@ -20,37 +18,49 @@ pub mod schema;
...
 
 pub struct Supergraph {
-    pub schema: Valid<Schema>,
+    pub schema: ValidFederationSchema,
 }
 
 impl Supergraph {
     pub fn new(schema_str: &str) -> Result<Self, FederationError> {
         let schema = Schema::parse_and_validate(schema_str, "schema.graphql")?;
         // TODO: federation-specific validation
```
SimonSapin, line 32: Is this TODO still relevant?

src/lib.rs
```diff
@@ -141,7 +151,7 @@ mod tests {
         "#;
 
         let supergraph = Supergraph::new(schema).unwrap();
-        let _subgraphs = database::extract_subgraphs(&supergraph)
+        let _subgraphs = crate::query_graph::extract_subgraphs_from_supergraph::extract_subgraphs_from_supergraph(&supergraph.schema, None)
             .expect("Should have been able to extract subgraphs");
         // TODO: actual assertions on the subgraph once it's actually implemented.
```
SimonSapin, line 156: Is it time to add assertions there? Does JS have corresponding tests?

apollo-bot2:
## Detected SAST Vulnerabilities

src/query_plan/operation.rs
```diff
@@ -1690,7 +1690,8 @@ pub(crate) fn equal_selection_sets(
     // from the JS codebase. It may be more performant for federation-next to use its own
     // representation instead of repeatedly inter-converting between its representation and the
     // apollo-rs one, but we'll cross that bridge if we come to it.
-    todo!();
+    println!("TODO: equal_selection_sets");
```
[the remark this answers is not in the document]
goto-bus-stop: Removing in favour of `==`

src/lib.rs
goto-bus-stop: AFAICT this test is not directly in JS, but we could add some that just assert the whole subgraph shape is as expected here.

src/query_plan/operation.rs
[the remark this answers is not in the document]
goto-bus-stop: I looked at where the conditions come from and I'm pretty sure it's important that the same selections from different subgraphs compare equal, so the only usage site of `equal_selection_sets` has to become `conditions.selections == followup_conditions.selections`. Done here https://github.com/apollographql/federation-next/pull/175/commits/35bc00a749e28ccde5fac879fd74c605112df13c

src/query_plan/query_planner.rs
```diff
@@ -106,26 +120,303 @@ impl Default for QueryPlannerDebugConfig {
...
+                        .get(&position.type_name)
+                        .is_some_and(is_interface_object)
+                })
+            })
+            .collect::<IndexSet<_>>();
+
+        let is_inconsistent = |position: AbstractTypeDefinitionPosition| {
+            let mut sources = query_graph.sources().filter_map(|subgraph| {
+                match subgraph.try_get_type(position.type_name().clone())? {
+                    TypeDefinitionPosition::Object(object) => Some(IndexSet::from([object])),
```
SimonSapin, line 200: The JS code skips object types (with a comment explaining why). Is there a reason to do something different here?

src/query_plan/query_planner.rs
```diff
@@ -106,26 +120,303 @@ impl Default for QueryPlannerDebugConfig {
...
+            .schema
+            .get_types()
+            .filter_map(|position| AbstractTypeDefinitionPosition::try_from(position).ok())
+            .filter(|position| is_inconsistent(position.clone()))
+            .collect::<IndexSet<_>>();
+
+        // TODO(@goto-bus-stop): not sure if this is needed
+        // we can just do `.get().unwrap_or(false)` on a map with *some* labels,
+        // instead of preparing this map for all labels?
+        let _default_override_conditions = supergraph
```
SimonSapin, line 237: I think I agree. And `BuildQueryPlanOptions.overrideConditions` could be a set of strings (where presence means true) instead of a map of string to booleans

SimonSapin (approved):

PR #177, opened 2024-02-14
chore(deps): update apollo-compiler to beta.13

goto-bus-stop:
Nothing particularly interesting for fed-next in this release. But after upgrading & releasing fed-next we can start testing API schema in router.


lrlna (approved):
ahhh so excited for API schema to go out!!

PR #220, opened 2024-02-20
@link(as: String) argument is nullable

SimonSapin:
Per spec: https://specs.apollo.dev/link/v1.0/#@link.as

### DIFF
--- a/src/subgraph/spec.rs
+++ b/src/subgraph/spec.rs
@@ -643,8 +643,7 @@ impl LinkSpecDefinitions {
                 InputValueDefinition {
                     description: None,
                     name: name!("url"),
-                    // TODO: doc-comment disagrees with non-null here
-                    ty: ty!(String!).into(),
+                    ty: ty!(String).into(),
                     default_value: None,
                     directives: Default::default(),
                 }


src/subgraph/spec.rs
```diff
@@ -643,8 +643,7 @@ impl LinkSpecDefinitions {
                 InputValueDefinition {
                     description: None,
                     name: name!("url"),
-                    // TODO: doc-comment disagrees with non-null here
-                    ty: ty!(String!).into(),
+                    ty: ty!(String).into(),
```
goto-bus-stop, line 646: this is the definition for `url`, which by spec is `String!`

dariuszkuc: guessing the confusion is from the comment on the method?

```
    ///   directive @link(url: String, as: String, import: [Import], for: link__Purpose) repeatable on SCHEMA
    pub fn link_directive_definition(&self) -> Result<DirectiveDefinition, FederationSpecError> {
```

Yeah it should be `url: String!` as it is a required param -> https://specs.apollo.dev/link/v1.0/

goto-bus-stop (approved):

PR #223, opened 2024-02-22
feat: support legacy `@core` link syntax

goto-bus-stop:
This is for backward compatibility with existing supergraphs, bolting `@core` support onto the `@link` reading. It does not do anything to support fed v1 supergraphs.

While `@core` is not used in new graphs today, it is supported by the JS code, and we have a lot of tests throughout our systems that rely on using `@core` with fed v2.

Additionally, because the implementation of API schema generation is the same between fed v1 and fed v2, supporting `@core` will let us land it in the router *without* a fallback to JS for fed v1 supergraphs.

All new code that supports `@core` mentions `@core` literally in a comment so we can rip it out easily when we migrate over customers and our tests in the router.

### DIFF
--- a/src/link/database.rs
+++ b/src/link/database.rs
@@ -15,7 +15,7 @@ pub fn links_metadata(schema: &Schema) -> Result<Option<LinksMetadata>, LinkErro
         .schema_definition
         .directives
         .iter()
-        .filter(|d| parse_link_if_bootstrap_directive(schema, d));
+        .filter(|d| is_bootstrap_directive(schema, d));
     let Some(bootstrap_directive) = bootstrap_directives.next() else {
         return Ok(None);
     };
@@ -163,9 +163,9 @@ fn is_core_directive_definition(definition: &DirectiveDefinition) -> bool {
             .map_or(true, |argument| *argument.ty == ty!(String))
 }
 
-// Note: currently only recognizing @link, not @core. Doesn't feel worth bothering with @core at
-// this point, but the latter uses the "feature" arg instead of "url".
-fn parse_link_if_bootstrap_directive(schema: &Schema, directive: &Directive) -> bool {
+/// Returns whether a given directive is the @link or @core directive that imports the @link or
+/// @core spec.
+fn is_bootstrap_directive(schema: &Schema, directive: &Directive) -> bool {
     let Some(definition) = schema.directive_definitions.get(&directive.name) else {
         return false;
     };
--- a/src/link/database.rs
+++ b/src/link/database.rs
@@ -11,6 +11,9 @@ use crate::link::{
 
 /// Extract @link metadata from a schema.
 pub fn links_metadata(schema: &Schema) -> Result<Option<LinksMetadata>, LinkError> {
+    // This finds "bootstrap" uses of @link / @core regardless of order. By spec,
+    // the bootstrap directive application must be the first application of @link / @core, but
+    // this was not enforced by the JS implementation, so we match it for backward compatibility.
     let mut bootstrap_directives = schema
         .schema_definition
         .directives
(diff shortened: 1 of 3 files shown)


src/link/database.rs
```diff
@@ -191,6 +244,74 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn renamed_link_directive() -> Result<(), LinkError> {
+        let schema = r#"
+          extend schema
+            @lonk(url: "https://specs.apollo.dev/link/v1.0", as: "lonk")
```
goto-bus-stop, line 251: Didn't see any tests for the renaming yet, hence

src/link/database.rs
```diff
@@ -106,31 +114,66 @@ pub fn links_metadata(schema: &Schema) -> Result<Option<LinksMetadata>, LinkErro
...
+/// directive @_ANY_NAME_(feature: String!) repeatable on SCHEMA
+/// directive @_ANY_NAME_(feature: String) repeatable on SCHEMA
+/// ```
+fn is_core_directive_definition(definition: &DirectiveDefinition) -> bool {
+    // XXX(@goto-bus-stop): @core compatibility is primarily to support old tests--should be
+    // removed when those are updated.
+    definition.repeatable
+        && definition.locations == [DirectiveLocation::Schema]
+        && definition
+            .argument_by_name("feature")
```
SimonSapin, line 155: Shouldn’t this also check that `defintion` does not have unexpected arguments?

src/link/database.rs
```diff
@@ -106,31 +114,66 @@ pub fn links_metadata(schema: &Schema) -> Result<Option<LinksMetadata>, LinkErro
...
+///
+/// Either of these definitions are accepted:
+/// ```graphql
+/// directive @_ANY_NAME_(url: String!, as: String) repeatable on SCHEMA
+/// directive @_ANY_NAME_(url: String, as: String) repeatable on SCHEMA
+/// ```
+fn is_link_directive_definition(definition: &DirectiveDefinition) -> bool {
+    definition.repeatable
+        && definition.locations == [DirectiveLocation::Schema]
+        && definition.argument_by_name("url").is_some_and(|argument| {
```
SimonSapin, line 127: Shouldn’t this also check that `defintion` does not have unexpected arguments?

src/link/database.rs
```diff
@@ -1,33 +1,41 @@
...
+                // XXX(@goto-bus-stop): @core compatibility is primarily to support old tests--should be
+                // removed when those are updated.
+                .or(extraneous_directive.argument_by_name("feature"))
+                .and_then(|value| value.as_str().map(Cow::Borrowed))
+                .unwrap_or_else(|| Cow::Owned(Identity::link_identity().to_string()))
         )));
     }
+
     // At this point, we know this schema uses "our" @link. So we now "just" want to validate
     // all of the @link usages (starting with the bootstrapping one) and extract their metadata.
```
SimonSapin, line 37: https://specs.apollo.dev/link/v1.0/ does not have a corresponding section but https://specs.apollo.dev/core/v0.2/#@core/feature has an extra check that the boostrap directive should be the first one listed with that name.

src/link/database.rs
goto-bus-stop: I would agree, though the JS code doesn't appear to check for this, maybe for forward compatibility..? I think that seems risky since the interpretation would likely be wrong if there's an unsupported argument.

src/link/database.rs
goto-bus-stop: Yeah it's a good point. JS doesn't check this, I would be surprised if there are any schemas where it's not true (since supergraphs are virtually always generated) but I don't really want to risk that incompatibility. I added a comment about it at the top of the function where we look up the bootstrap directive.

src/link/database.rs
dariuszkuc: we should keep it the same as the JS code

goto-bus-stop:
Marking the comments resolved because we'll follow the JS code rather than the spec (or good sense 😄 ). this just needs a rubber-stamp approval I'd say

dariuszkuc (approved):

Issue #158, opened 2024-01-26
Port FetchDependencyGraphNode::to_plan_node


SimonSapin:
Following up https://github.com/apollographql/federation-next/pull/121, port [FetchGroup.toPlanNode](https://github.com/apollographql/federation/blob/25beb382fff253ac38ef6d7a5454af60da0addbb/query-planner-js/src/buildPlan.ts#L1488) to `FetchDependencyGraphNode::to_plan_node`

PR #226, opened 2024-03-01
[Plan generation] Port FetchGroup.toPlanNode

SimonSapin:
Fixes https://github.com/apollographql/federation-next/issues/158

### DIFF
--- a/src/query_plan/fetch_dependency_graph_processor.rs
+++ b/src/query_plan/fetch_dependency_graph_processor.rs
@@ -1,9 +1,10 @@
 use crate::error::FederationError;
 use crate::query_plan::conditions::Conditions;
 use crate::query_plan::fetch_dependency_graph::DeferredInfo;
+use crate::query_plan::fetch_dependency_graph::FetchDependencyGraph;
 use crate::query_plan::fetch_dependency_graph::FetchDependencyGraphNode;
+use crate::query_plan::operation::NamedFragments;
 use crate::query_plan::operation::NormalizedSelectionSet;
-use crate::query_plan::query_planner::QueryPlannerConfig;
 use crate::query_plan::ConditionNode;
 use crate::query_plan::DeferNode;
 use crate::query_plan::DeferredDeferBlock;
@@ -13,10 +14,12 @@ use crate::query_plan::PlanNode;
 use crate::query_plan::PrimaryDeferBlock;
 use crate::query_plan::QueryPlanCost;
 use crate::query_plan::SequenceNode;
+use crate::schema::ValidFederationSchema;
 use apollo_compiler::ast::Name;
 use apollo_compiler::executable::VariableDefinition;
 use apollo_compiler::Node;
 use apollo_compiler::NodeStr;
+use std::collections::HashMap;
 use std::collections::HashSet;
 use std::sync::Arc;
 
@@ -43,7 +46,6 @@ const FETCH_COST: QueryPlanCost = 1000;
 const PIPELINING_COST: QueryPlanCost = 100;
 
 pub(crate) struct FetchDependencyGraphToQueryPlanProcessor {
-    config: Arc<QueryPlannerConfig>,
     variable_definitions: Vec<Node<VariableDefinition>>,
     fragments: Option<RebasedFragments>,
     operation_name: Option<Name>,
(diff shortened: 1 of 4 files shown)


src/query_plan/fetch_dependency_graph_processor.rs
```diff
@@ -413,3 +419,30 @@ fn flat_wrap_nodes(
...
+            .entry(subgraph_name)
+            .or_insert_with(|| Self::rebase_on(&self.query_fragments, subgraph_schema))
+            .as_ref()
+    }
+
+    fn rebase_on(
+        query_fragments: &NamedFragments,
+        subgraph_schema: &ValidFederationSchema,
+    ) -> Option<NamedFragments> {
+        todo!() // TODO: port JS `NamedFragments.rebaseOn` from `operations.ts`
```
Geal, line 446: this will be implemented in https://github.com/apollographql/federation-next/pull/239

SimonSapin (commented):
I’m happy to merge this now, but since I’m making a non-trivial amount of new changes it’d be good for those to get review too.

src/query_plan/fetch_dependency_graph_processor.rs
```diff
@@ -89,6 +87,7 @@ pub(crate) struct FetchDependencyGraphToCostProcessor;
 pub(crate) trait FetchDependencyGraphProcessor<TProcessed, TDeferred> {
     fn on_node(
         &mut self,
+        graph: &FetchDependencyGraph,
         node: &mut FetchDependencyGraphNode,
```
goto-bus-stop, line 91: Nodes are part of a FetchDependencyGraph, so this API requires a mutable reference and an immutable reference to be alive at the same time. we should provide a graph and a NodeIndex instead of a mutable reference to the node (this will cause a small conflict with #261).

src/query_plan/operation.rs
```diff
@@ -1043,6 +1182,13 @@ impl NormalizedFragmentSpreadSelection {
         })
     }
 
+    pub(crate) fn with_updated_selection_set(
+        &self,
+        _selection_set: Option<NormalizedSelectionSet>,
+    ) -> Self {
+        unimplemented!("unsupported")
```
goto-bus-stop, line 1189:
```suggestion
        todo!("unsupported")
```
since we use `todo!()` everywhere else, can we do so here so there's only one thing to grep for? or is it intentionally different?

src/query_plan/operation.rs
```diff
@@ -1125,6 +1275,27 @@ pub(crate) mod normalized_inline_fragment_selection {
...
     }
 
+    impl NormalizedInlineFragmentSelection {
+        pub(crate) fn with_updated_selection_set(
+            &self,
+            selection_set: Option<NormalizedSelectionSet>,
+        ) -> Self {
+            Self {
+                inline_fragment: self.inline_fragment.clone(),
+                //FIXME
```
goto-bus-stop, line 1285: What needs to be FIXME'd here?

goto-bus-stop (commented):
@SimonSapin LGTM at first glance aside from the things I commented, I didn't extensively compare it to the JS codebase yet though. Would you like me to?

src/query_plan/operation.rs
SimonSapin: I guess the use of `unwrap`? Not sure what the best fix would be though. In the one (eventual) call site, `None` is only passed here if `NormalizedSelection::selection_set` returned `None` for the same selection.

src/query_plan/operation.rs
```diff
@@ -747,6 +809,43 @@ pub(crate) mod normalized_field_selection {
...
+
+        pub(crate) fn element(&self) -> OpPathElement {
+            OpPathElement::Field(self.field.clone())
+        }
+
+        pub(crate) fn with_updated_alias(&self, alias: Name) -> NormalizedField {
+            let mut data = self.field.data().clone();
+            data.alias = Some(alias);
+            NormalizedField::new(data)
+            // FIXME: what is copyAttachments?
```
goto-bus-stop, line 831: "attachments" in JS is `sibling_typename` in rust.

src/query_plan/fetch_dependency_graph_processor.rs
SimonSapin: I’ve changed the parameter to only `&QueryGraph`. It should be possible to borrow disjoint fields of `FederatedDepnedencyGraph` at the same time.

src/query_plan/operation.rs
SimonSapin: Alright, so it’s already part of `NormalizedFieldData` and nothing else is needed here.

SimonSapin:
> I didn't extensively compare it to the JS codebase

I did, so I think we’re good if you’ve seen my recent changes

goto-bus-stop (approved):

PR #229, opened 2024-03-12
[Query Planner] add stub implementation for QueryPlanner::build_query_plan

goto-bus-stop:
Just call into the parts that already exist (operation normalisation)

This is intended to be a stub to start experimenting with router integration (like getting all the types in order).

### DIFF
--- a/src/query_plan/query_planner.rs
+++ b/src/query_plan/query_planner.rs
@@ -438,9 +438,9 @@ type User
     #[test]
     fn it_does_not_crash() {
         let supergraph = Supergraph::new(TEST_SUPERGRAPH).unwrap();
-        let planner = QueryPlanner::new(&supergraph, Default::default()).unwrap();
+        let _planner = QueryPlanner::new(&supergraph, Default::default()).unwrap();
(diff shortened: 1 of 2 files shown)


src/query_plan/query_planner.rs
```diff
@@ -140,11 +146,12 @@ pub struct QueryPlanner {
     abstract_types_with_inconsistent_runtime_types: IndexSet<AbstractTypeDefinitionPosition>,
     // TODO: Port _lastGeneratedPlanStatistics from the JS codebase in a way that keeps QueryPlanner
     // immutable.
+    // last_generated_plan_statistics: QueryPlanningStatistics,
 }
 
 impl QueryPlanner {
     pub fn new(
-        supergraph: Supergraph,
+        supergraph: &Supergraph,
```
goto-bus-stop, line 154: This doesn't need ownership as it just copies the things it needs off the supergraph into the query planner struct.

lrlna (approved):

SimonSapin (approved):

PR #231, opened 2024-03-14
Port `ExternalTester` from the JS codebase

sachindshinde:
This PR:
- Adds subgraph metadata (known as `FederationMetadata` in the JS codebase) to the schema representation, but doesn't port all the logic there.
- Ports external metadata (known as `ExternalTester` in the JS codebase).


goto-bus-stop (changes requested):
Implementation looks correct, just some Rusty comments!

src/schema/mod.rs
```diff
@@ -180,42 +183,81 @@ impl FederationSchema {
 
 /// A GraphQL schema with federation data that is known to be valid, and cheap to clone.
 #[derive(Debug, Clone)]
-pub struct ValidFederationSchema(pub(crate) Arc<Valid<FederationSchema>>);
+pub struct ValidFederationSchema {
+    schema: Arc<Valid<FederationSchema>>,
+    subgraph_metadata: Option<SubgraphMetadata>,
```
goto-bus-stop, line 188: Oh, this is also an issue, `SubgraphMetadata` is not cheap to clone but is part of `ValidFederationSchema` which (in theory) is.

I think we need to tweak the order of initialisation to fix this and maybe let SubgraphMetadata work on not-necessarily-valid federation schemas.

goto-bus-stop:
I hope it's okay that I addressed my own feedback, that's how I interpreted it in the meeting yesterday, and I was looking for something fun to do!

goto-bus-stop:
`FederationSchema` is used for supergraphs and subgraphs, maybe this should be included in a wrapper type that's only for subgraphs?

goto-bus-stop:
I moved `SubgraphMetadata` to be part of `FederationSchema` instead of `ValidFederationSchema`, but it's only populated in the `ValidFederationSchema` constructor and it can only be accessed through a `ValidFederationSchema`.

I think we should have separate types for subgraph and supergraph schemas, since the code always knows which one it is operating on, and they require different data and different handling.

lrlna (approved):

PR #234, opened 2024-03-19
implement removeConditionsFromSelectionSet

Geal:
Note: all uses or FieldDefinitionPosition are probably wrong, because it does not take into account NonNull or List wrappers


apollo-bot2:
## Detected SAST Vulnerabilities

duckki:
I'm curious what SAST Vulnerabilities were detected. Is it a false alarm?

Geal:
There was no context at all on what was detected

PR #238, opened 2024-03-25
Derive PartialEq and Hash for OpGraphPathContext

TylerBloom:
This PR addresses Jira ticket FED-28.

After verifying that the `PartialEq` implementation on `Arc` does the desired short-circuiting and the derived behavior matches what is expected in the JS code, the stubs for the manual implementation of `PartialEq` and `Hash` were replaced with derived implementations.


dariuszkuc (approved):

goto-bus-stop:
@TylerBloom I think you need to log in to CircleCI to get the remaining jobs to run

PR #247, opened 2024-04-01
Ported the GraphViz `dot`-format output for query graphs

duckki:
added query_graph::output::to_dot function
- outputs query graphs in the GraphViz Dot format
- supports both federated schemas and non-federated ones

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -80,7 +80,7 @@ fn load_supergraph(
             .iter()
             .map(|pathname| {
                 let doc_str = std::fs::read_to_string(pathname).unwrap();
-                let url = "file://".to_string() + pathname.to_str().unwrap();
+                let url = format!("file://{}", pathname.to_str().unwrap());
                 let basename = pathname.file_stem().unwrap().to_str().unwrap();
                 subgraph::Subgraph::parse_and_expand(basename, &url, &doc_str).unwrap()
             })
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -37,27 +37,24 @@ enum Command {
 fn main() -> ExitCode {
     let args = Args::parse();
     match args.command {
-        Command::Api { supegraph_schema } =>
-            run_federation_task( &|| { to_api_schema(&supegraph_schema) } ),
-        Command::QueryGraph { schemas } => {
-            run_federation_task( &|| { dot_query_graph(&schemas) } )
-        },
+        Command::Api { supegraph_schema } => {
+            run_federation_task(&|| to_api_schema(&supegraph_schema))
+        }
+        Command::QueryGraph { schemas } => run_federation_task(&|| dot_query_graph(&schemas)),
         Command::FederatedGraph { schemas } => {
-            run_federation_task( &|| { dot_federated_graph(&schemas) } )
-        },
+            run_federation_task(&|| dot_federated_graph(&schemas))
+        }
     }
 }
 
-fn run_federation_task( func: &dyn Fn() -> Result<(), FederationError> ) -> ExitCode {
+fn run_federation_task(func: &dyn Fn() -> Result<(), FederationError>) -> ExitCode {
     match func() {
         Err(error) => {
             eprintln!("{error}");
             ExitCode::FAILURE
-        },
-
-        Ok(_) => {
-            ExitCode::SUCCESS
         }
+
+        Ok(_) => ExitCode::SUCCESS,
     }
 }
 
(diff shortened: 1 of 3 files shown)


duckki:
I didn't add CLI, yet. But, I can added the `querygraph` and `federatedgraph` CLI commands once https://github.com/apollographql/federation-next/pull/246 merges.

goto-bus-stop (commented):
Nice, implementation looks good. I think most of the `dot_str.push_str(&format!())` calls would read better with `write!(&mut dot_str, etc)` (or `writeln!()` as appropriate). I have some rust comments for the CLI part.

cli/src/main.rs
```diff
@@ -17,27 +22,98 @@ enum Command {
...
+    if file_paths.len() == 0 {
+        panic!("Error: missing command arguments");
+    } else if file_paths.len() == 1 {
+        let doc_str = std::fs::read_to_string(&file_paths[0]).unwrap();
+        apollo_federation::Supergraph::new(&doc_str)
+    } else {
+        let schemas: Vec<_> = file_paths
+            .iter()
+            .map(|pathname| {
+                let doc_str = std::fs::read_to_string(&pathname).unwrap();
```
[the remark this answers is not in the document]
duckki: I don't think it's necessary for this CLI. But, I'll look into anyhow and use it where ever appropriate!

duckki:
@goto-bus-stop Thanks for the review! I just pushed a new commit. Please review one more time.

src/query_graph/output.rs
```diff
@@ -0,0 +1,149 @@
...
+            &cluster_dot_config,
+            &(|_, er| label_edge(er.weight())),
+            &(|_, (_, node)| label_cluster_node(node)),
+        )
+        .to_string();
+
+        writeln!(dot_str, r#"  subgraph "cluster_{}" {{"#, &cluster_name)?;
+        writeln!(dot_str, r#"    label = "Subgraph \"{}\"";"#, &cluster_name)?;
+        writeln!(dot_str, r#"    color = "black";"#)?;
+        writeln!(dot_str, r#"    style = """#)?;
```
goto-bus-stop, line 118: I'm not too familiar with dot syntax, but does this need a semicolon?
```suggestion
        writeln!(dot_str, r#"    style = "";"#)?;
```

cli/src/main.rs
```diff
@@ -17,27 +22,91 @@ enum Command {
...
+        panic!("Error: missing command arguments");
+    } else if file_paths.len() == 1 {
+        let doc_str = std::fs::read_to_string(&file_paths[0]).unwrap();
+        apollo_federation::Supergraph::new(&doc_str)
+    } else {
+        let schemas: Vec<_> = file_paths
+            .iter()
+            .map(|pathname| {
+                let doc_str = std::fs::read_to_string(pathname).unwrap();
+                let url = "file://".to_string() + pathname.to_str().unwrap();
```
goto-bus-stop, line 83: Ah, missed this the first time, it's almost always better to use the `format!()` macro than the string `+` operator, it reads a bit better and requires fewer memory allocations.
```suggestion
                let url = format!("file://{}", pathname.to_str().unwrap());
```

src/query_graph/output.rs
```diff
@@ -0,0 +1,149 @@
...
+        cluster_name: &NodeStr,
+        edge_index: EdgeIndex,
+    ) -> bool {
+        match graph.edge_endpoints(edge_index) {
+            None => false,
+
+            Some((n1, n2)) => {
+                graph[n1].source == *cluster_name && graph[n2].source == *cluster_name
+            }
+        }
```
goto-bus-stop, line 59: This is perfectly fine but you can also use the [`.is_some_and`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.is_some_and) method here:
```suggestion
        graph.edge_endpoints(edge_index).is_some_and(|(n1, n2)| {
            graph[n1].source == *cluster_name && graph[n2].source == *cluster_name
        })
```

src/query_graph/output.rs
goto-bus-stop: Ah, it looks like it's optional.

goto-bus-stop (approved):
LGTM, some more stylistic nits if you feel like it, but no worries :)

src/query_graph/output.rs
duckki: I believe so. I tried mimic the JS implementation's behavior :)

src/query_graph/output.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn edge_across_clusters(graph: &StableInnerGraph, edge_index: EdgeIndex) -> bool {
+        match graph.edge_endpoints(edge_index) {
+            None => false,
+
+            Some((n1, n2)) => graph[n1].source != graph[n2].source,
+        }
+    }
+
+    fn label_cluster_node(node: &QueryGraphNode) -> String {
+        format!(r#"label="{}@{}""#, node.type_, node.source)
```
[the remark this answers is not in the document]
duckki: makes sense. will do!

PR #248, opened 2024-04-02
Ported maybe_dump_subgraph_schema from JS

TylerBloom:
This PR addresses Jira ticket FED-53.

This method, ported from JS, appends additional information to error messages when attempting to extract subgraphs from a supergraph. Users can also set the environment variable `APOLLO_FEDERATION_DEBUG_SUBGRAPHS` to have the schema written to a file.

### DIFF
--- a/src/query_graph/extract_subgraphs_from_supergraph.rs
+++ b/src/query_graph/extract_subgraphs_from_supergraph.rs
@@ -32,6 +32,7 @@ use indexmap::{IndexMap, IndexSet};
 use lazy_static::lazy_static;
 use std::collections::BTreeMap;
 use std::fmt;
+use std::fmt::Write;
 use std::ops::Deref;
 use std::sync::Arc;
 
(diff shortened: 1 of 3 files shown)


TylerBloom:
After looking at the dependency differences, I decided to add `chrono` with all of the default features disabled except `clock` (to get the local time). This added 4 transitive dependencies, bringing the total number of deps from 71 to 76. Surprisingly, this is one fewer transitive deps than the minimum config for the `time` crate.

dariuszkuc:
> After looking at the dependency differences, I decided to add `chrono` with all of the default features disabled except `clock` (to get the local time). This added 4 transitive dependencies, bringing the total number of deps from 71 to 76. Surprisingly, this is one fewer transitive deps than the minimum config for the `time` crate.

@TylerBloom I believe the reason why `time` was mentioned is that this was already `apollographql/router` dependency so   we wouldn't be adding any new dependencies (assuming we would keep those versions in sync).

src/query_graph/extract_subgraphs_from_supergraph.rs
```diff
@@ -2036,3 +2036,35 @@ fn is_external_or_has_external_implementations(
...
     Ok(false)
 }
+
+static DEBUG_SUBGRAPHS_ENV_VARIABLE_NAME: &str = "APOLLO_FEDERATION_DEBUG_SUBGRAPHS";
+
+fn maybe_dump_subgraph_schema(subgraph: FederationSubgraph, message: &mut String) {
+    // NOTE: The std::fmt::write returns an error, but writing to a string will never return an
+    // error, so it is dropped when called.
+
+    if let Some(true) = std::env::var(DEBUG_SUBGRAPHS_ENV_VARIABLE_NAME)
```
dariuszkuc, line 2046: I think this warning message should be printed for false/undefined case? In JS we do `!!<envBoolValue>` which also correctly handles `undefined`

dariuszkuc (changes requested):

src/query_graph/extract_subgraphs_from_supergraph.rs
TylerBloom: Oops, you're right. Addressed it in 4e7d7c9b59875f3a1f97feb657de858ba28d9f78

dariuszkuc (approved):
NIT: Since it looks somewhat odd to match on true to just do nothing, I'd probably just do `if let Some(true) = match .... { // dump file } else { // print warning }`

PR #249, opened 2024-04-03
Implemented `Display` trait for OpPathTree

duckki:
#### Side note
The original suggestion was to implement the trait for `PathTree` type. But, it would be messy to define all the necessary trait constraints for the type parameters. So, I did it for `OpPathTree`, instead.

### DIFF
--- a/src/query_graph/path_tree.rs
+++ b/src/query_graph/path_tree.rs
@@ -356,3 +356,165 @@ fn merge_conditions(
         (None, None) => None,
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+
+    use apollo_compiler::{executable::DirectiveList, ExecutableDocument};
+    use petgraph::{stable_graph::NodeIndex, visit::EdgeRef};
+
+    use crate::{
+        error::FederationError,
+        query_graph::{
+            build_query_graph::build_query_graph,
+            condition_resolver::ConditionResolution,
+            graph_path::{OpGraphPath, OpGraphPathTrigger, OpPathElement},
+            path_tree::OpPathTree,
+            QueryGraph, QueryGraphEdgeTransition,
+        },
+        query_plan::operation::{
+            normalize_operation,
+            normalized_field_selection::{NormalizedField, NormalizedFieldData},
+        },
+        schema::{position::SchemaRootDefinitionKind, ValidFederationSchema},
+    };
+
+    // NB: stole from operation.rs
+    fn parse_schema_and_operation(
+        schema_and_operation: &str,
+    ) -> (ValidFederationSchema, ExecutableDocument) {
+        let (schema, executable_document) =
+            apollo_compiler::parse_mixed_validate(schema_and_operation, "document.graphql")
+                .unwrap();
+        let executable_document = executable_document.into_inner();
+        let schema = ValidFederationSchema::new(schema).unwrap();
+        (schema, executable_document)
+    }
+
+    fn trivial_condition() -> ConditionResolution {
+        ConditionResolution::Satisfied {
+            cost: 0,
+            path_tree: None,
+        }
+    }
+
+    // A helper function that builds a graph path from a sequence of field names
+    fn build_graph_path(
+        query_graph: &Arc<QueryGraph>,
+        op_kind: SchemaRootDefinitionKind,
+        path: &[&str],
+    ) -> Result<OpGraphPath, FederationError> {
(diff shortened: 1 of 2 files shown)


src/query_graph/path_tree.rs
```diff
@@ -111,13 +111,59 @@ impl OpPathTree {
...
     }
+
+    fn fmt_internal(
+        &self,
+        f: &mut Formatter<'_>,
+        indent: &String,
+        include_conditions: bool,
+    ) -> std::fmt::Result {
+        if self.is_leaf() {
+            return writeln!(f, "{}", self.vertex());
```
goto-bus-stop, line 122: I think in this case we should not add a newline, as it's a one-line path string? The JS code doesn't add a newline here and it makes sense for the caller to decide if it should be followed by a newline or not.
```suggestion
            return write!(f, "{}", self.vertex());
```

src/query_graph/path_tree.rs
```diff
@@ -111,13 +111,59 @@ impl OpPathTree {
         }
         Ok(true)
     }
+
+    fn fmt_internal(
+        &self,
+        f: &mut Formatter<'_>,
+        indent: &String,
```
goto-bus-stop, line 118:
```suggestion
        indent: &str,
```
Should take a `&str` instead of an owned String whenever possible, it's more flexible and `String` easily dereferences to `&str`.

src/query_graph/path_tree.rs
```diff
@@ -111,13 +111,59 @@ impl OpPathTree {
...
+        include_conditions: bool,
+    ) -> std::fmt::Result {
+        if self.is_leaf() {
+            return writeln!(f, "{}", self.vertex());
+        }
+        writeln!(f, "{}:", self.vertex())?;
+        let child_indent = indent.to_owned() + "  ";
```
goto-bus-stop, line 125: A little prettier:
```suggestion
        let child_indent = format!("{indent}  ");
```

src/query_graph/path_tree.rs
```diff
@@ -111,13 +111,59 @@ impl OpPathTree {
...
+        for child in self.childs.iter() {
+            let index = match child.edge {
+                Some(index) => index,
+                None => EdgeIndex::end(),
+            };
```
goto-bus-stop, line 130: There's a function for that! 😄 
```suggestion
            let index = child.edge.unwrap_or_else(EdgeIndex::end);
```

src/query_graph/path_tree.rs
```diff
@@ -111,13 +111,59 @@ impl OpPathTree {
...
+    }
 }
 
 impl<TTrigger, TEdge> PathTree<TTrigger, TEdge>
 where
     TTrigger: Eq + Hash,
     TEdge: Copy + Hash + Eq + Into<Option<EdgeIndex>>,
 {
+    /// Returns the `QueryGraphNode` represented by `self.node`.
+    /// - Note: This is named after the JS implementation's `vertex` field.
```
goto-bus-stop, line 161: We conventionally use `PORT_NOTE` to point out differences between the implementations, I'm not sure which of the two fits the best, but as long as we say `PORT_NOTE` or `TODO` we will be able to easily find places like this to revisit later.
```suggestion
    /// PORT_NOTE: This is named after the JS implementation's `vertex` field.
```

goto-bus-stop (approved):
Looks great! I think it's a good decision to put this on `OpPathTree` and save ourselves the generics. (We might not need the PathTree / OpPathTree difference at all. The only types in the JS codebase are `OpPathTree` and `RootPathTree`, a difference that is currently decided by a runtime boolean instead of a separate type in the Rust codebase)

Only minor comments / nits below

duckki:
@goto-bus-stop Thanks, again! 👍

src/query_graph/path_tree.rs
```diff
@@ -356,3 +356,165 @@ fn merge_conditions(
...
+            QueryGraph, QueryGraphEdgeTransition,
+        },
+        query_plan::operation::{
+            normalize_operation,
+            normalized_field_selection::{NormalizedField, NormalizedFieldData},
+        },
+        schema::{position::SchemaRootDefinitionKind, ValidFederationSchema},
+    };
+
+    // NB: stole from operation.rs
```
dariuszkuc, line 380: whats NB?

duckki: Yes. It was pretty widely used back at Synopsys. Maybe because it's shorter than "Note", I think.

PR #250, opened 2024-04-04
added `validate`, `compose` and `extract` CLI commands

duckki:
- updated command arguments to be consistent across all commands

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1,7 +1,7 @@
 use clap::Parser;
 use std::fs;
 use std::io;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process::ExitCode;
 
 use apollo_compiler::ExecutableDocument;
@@ -72,7 +72,7 @@ fn main() -> ExitCode {
         Command::Extract {
             supergraph_schema,
             destination_dir,
-        } => cmd_extract(&supergraph_schema, destination_dir),
+        } => cmd_extract(&supergraph_schema, destination_dir.as_ref()),
     };
     match result {
         Err(error) => {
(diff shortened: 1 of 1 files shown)


lrlna (approved):

cli/src/main.rs
```diff
@@ -110,3 +145,40 @@ fn dot_federated_graph(file_paths: &[PathBuf]) -> Result<(), FederationError> {
...
+        fs::create_dir_all(dest).map_err(|_| SingleFederationError::Internal {
+            message: "Error: directory creation failed".into(),
+        })?;
+        for (name, subgraph) in subgraphs {
+            let subgraph_path = dest.join(format!("{}.graphql", name));
+            fs::write(subgraph_path, subgraph.schema.schema().to_string()).map_err(|_| {
+                SingleFederationError::Internal {
+                    message: "Error: file output failed".into(),
+                }
+            })?;
```
goto-bus-stop, line 174: We have a constructor function for internal federation errors:
```suggestion
            fs::write(subgraph_path, subgraph.schema.schema().to_string()).map_err(|_| {
                FederationError::internal("Error: file output failed")
            })?;
```

goto-bus-stop (commented):
Small comments, nothing blocking. There's a merge conflict due to #244 🙈

cli/src/main.rs
duckki: Currently, `FederationError::internal` is `pub(crate)` and not visible to CLI. Shall we open it up to public?

duckki:
@goto-bus-stop Thanks for comments. I'll merge this for now. Let's discuss separately whether FederationError::internal should be public or not.

PR #251, opened 2024-04-05
parse_field_set: forbid aliases

duckki:
Added a check for the absence of aliases in field set directive arguments.

### Rationale

Federation-next uses the apollo-compiler to parse the field-set argument string as GraphQL code. Similarly, the JS federation uses graphql-js to parse it. Both GraphQL parsers allow aliases in it, while federation does not allow it. That’s why the JS federation looks for aliases after parsing it. I think it makes sense to do the same thing in the Rust implementation so we don’t need to make changes in the apollo-compiler itself.

### Side note

I did not attempt to make our error message to match exactly with the JS implementation. First of all, it's not necessary for unblocking QP work, while improving error messages will require further refactoring elsewhere. We might want to revisit the whole error message business later when we get to composition porting.

### DIFF
--- a/src/error/mod.rs
+++ b/src/error/mod.rs
@@ -2,6 +2,7 @@ use crate::subgraph::spec::FederationSpecError;
 use apollo_compiler::validation::DiagnosticList;
 use apollo_compiler::{ast::InvalidNameError, validation::WithErrors};
 use lazy_static::lazy_static;
+use std::cmp::Ordering;
 use std::fmt::{Display, Formatter, Write};
 
 // What we really needed here was the string representations in enum form, this isn't meant to
@@ -484,6 +485,22 @@ impl FederationError {
             message: message.into(),
         }
         .into()
+    }
+}
+
+// Converts a SingleFederationError array into `Result<(), FederationError>`.
+// The return value can be either Ok, Err with a SingleFederationError or MultipleFederationErrors,
+// depending on the number of errors in the input.
+pub(crate) fn check_federation_errors(
+    errors: &[SingleFederationError],
+) -> Result<(), FederationError> {
+    match errors.len().cmp(&1) {
+        Ordering::Less => Ok(()),
+        Ordering::Equal => Err(errors[0].clone().into()),
+        Ordering::Greater => Err(MultipleFederationErrors {
+            errors: errors.to_vec(),
+        }
+        .into()),
     }
 }
 
--- a/src/error/mod.rs
+++ b/src/error/mod.rs
@@ -488,10 +488,10 @@ impl FederationError {
     }
 }
 
-// Converts a SingleFederationError array into `Result<(), FederationError>`.
+// Converts a SingleFederationError slice into `Result<(), FederationError>`.
 // The return value can be either Ok, Err with a SingleFederationError or MultipleFederationErrors,
 // depending on the number of errors in the input.
-pub(crate) fn check_federation_errors(
+fn convert_single_errors_to_result(
     errors: &[SingleFederationError],
 ) -> Result<(), FederationError> {
     match errors.len().cmp(&1) {
@@ -501,6 +501,19 @@ pub(crate) fn check_federation_errors(
             errors: errors.to_vec(),
         }
         .into()),
+    }
+}
+
+impl FromIterator<SingleFederationError> for Result<(), FederationError> {
+    fn from_iter<T: IntoIterator<Item = SingleFederationError>>(iter: T) -> Self {
+        let errors: Vec<_> = iter.into_iter().collect();
+        convert_single_errors_to_result(&errors)
+    }
+}
+
+impl MultipleFederationErrors {
+    pub fn to_result(&self) -> Result<(), FederationError> {
+        convert_single_errors_to_result(&self.errors)
     }
 }
 
--- a/src/error/mod.rs
+++ b/src/error/mod.rs
@@ -428,6 +428,14 @@ impl From<DiagnosticList> for MultipleFederationErrors {
     }
 }
 
+impl FromIterator<SingleFederationError> for MultipleFederationErrors {
+    fn from_iter<T: IntoIterator<Item = SingleFederationError>>(iter: T) -> Self {
+        Self {
+            errors: iter.into_iter().collect(),
+        }
+    }
+}
+
 #[derive(Debug, Clone, thiserror::Error)]
 pub struct AggregateFederationError {
     pub code: String,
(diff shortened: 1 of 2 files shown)


src/error/mod.rs
```diff
@@ -487,6 +488,22 @@ impl FederationError {
     }
 }
 
+// Converts a SingleFederationError array into `Result<(), FederationError>`.
+// The return value can be either Ok, Err with a SingleFederationError or MultipleFederationErrors,
+// depending on the number of errors in the input.
+pub(crate) fn check_federation_errors(
```
lrlna, line 494: I feel like `check_federation_errors` isn't doing a check, but more of a conversion 🤔 . For these sorts of things, we usually like a [`Into`](https://doc.rust-lang.org/std/convert/trait.Into.html) or [`TryInto`](https://doc.rust-lang.org/std/convert/trait.TryInto.html). 

What do you think about something like `pub(crate) struct SingleFederationErrorList { inner: Vec<SingleFederationError> }` (and pub(crate) so it's just internal). We already have a [`DiagnosticList` above that does a similar conversion](https://github.com/apollographql/federation-next/pull/251/files#diff-f43f218cc5fee24623f17725c9bc2c541d64b2e9c1d8ca5345de67ef2a3b7392R469-R474). That way [in the field_set.rs file](https://github.com/apollographql/federation-next/pull/251/files#diff-87a8f5443445f15c22bbc48ca16e97c3d26ac790c690ee10ed2d4ed4620418f7R27) you can just do an `alias_errors.into()` as your return

SimonSapin:
Since `apollo_compiler::executable::FieldSet` exists specifically for federation directives and `apollo_parser::Parser::parse_selection_set` exists specifically for `FieldSet`, would it make sense to do this directly in parser or compiler?

Also, is there anything else that should not be accepted in a FieldSet? Directives, arguments, inline fragments, fragment spreads?

src/error/mod.rs
duckki: I agree `check_federation_errors` is misleading. It's indeed a conversion function.

Originally, I wanted to use `Vec<SingleFederationErrors>` to collect errors, which includes the `Ok` case when empty. We sometimes use `MultipleFederationErrors` as both `Ok`(when the internal array is empty) and `Err` cases temporarily. But, I think `Vec` is clearer since `MultipleFederationErrors` will be cast into `Err` by default (so you have to check before returning it).

In order to convert `Vec<SingleFederationErrors>` into `Result`, I tried something like `From<&[SingleFederationError]> for Result<(), FederationError>`. But, that wasn't allowed. That's why I ended up with `check_federation_errors` function.

Now I think of this. I just got a better idea. While I can't convert `Vec<SingleFederationErrors>` using `into()`, since I have `FromIterator<SingleFederationErrors>`, I can do this:

```rust
impl FromIterator<SingleFederationError> for Result<(), FederationError> {
    fn from_iter<T: IntoIterator<Item=SingleFederationError>>(iter: T) -> Self {
        let errors: Vec<_> = iter.into_iter().collect();
        match errors.len().cmp(&1) {
            Ordering::Less => Ok(()),
            Ordering::Equal => Err(errors[0].clone().into()),
            Ordering::Greater => Err(MultipleFederationErrors {
                errors: errors.to_vec(),
            }
            .into()),
        }
    }
}
```
I'll push this version.

duckki:
> Since `apollo_compiler::executable::FieldSet` exists specifically for federation directives and `apollo_parser::Parser::parse_selection_set` exists specifically for `FieldSet`, would it make sense to do this directly in parser or compiler?
> 
> Also, is there anything else that should not be accepted in a FieldSet? Directives, arguments, inline fragments, fragment spreads?

Good point. The JS implementation only checked aliasing. I'll look into the other cases.

BTW, we tried not to change apollo_parser, leaving it for generic GraphQL. Otherwise, we'll need to add parser options to limit them.

duckki:
> Good point. The JS implementation only checked aliasing. I'll look into the other cases.

I've investigated this and here are my findings:

The spec (https://www.apollographql.com/docs/federation/subgraph-spec) doesn't articulate what's not allowed. So, I think it's up to the implementation.

Our JS federation allows arguments and inline spreads (with type conditions). The parser allows named fragments, but the validation will error out if used (since fragments can't be defined in schemas).  On the other hand, it disallows aliases and directives explicltly. Interestingly, aliases are checked in `parseFieldSetArgument` function (thus, checked every time it's parsed), but the directive part is checked in `validateFieldSetSelections` function (only called during validation).

For QP porting purposes, I believe we can defer the rest of validation for a later time.

src/error/mod.rs
duckki: @lrlna I pushed a refactored version. Please review :)

src/error/mod.rs
```diff
@@ -487,6 +488,35 @@ impl FederationError {
     }
 }
 
+// Converts a SingleFederationError slice into `Result<(), FederationError>`.
+// The return value can be either Ok, Err with a SingleFederationError or MultipleFederationErrors,
+// depending on the number of errors in the input.
+fn convert_single_errors_to_result(
+    errors: &[SingleFederationError],
```
lrlna, line 495: usually in a function like this, you'd want to right away take ownership of the vector. we are converting from one type to another, so there isn't a reason to have this slice. otherwise you're doing an extra clone that isn't necessary.

src/error/mod.rs
lrlna: I think we can simplify this a bit. The `IntoIterator` is doing a bit more than it should here. It looks like we want to have two things: a `Vec<SingleFederationError>` to be converted to `FederationError` to then be converted to `Result<(), FederationError>`.

The most rust-y way of doing it is doing what we do with `DiagnosticList` in `apollo-compiler`. 

a `Vec<Diagnostic>` gets converted to `DiagnosticList` (with an `Into` impl), a `DiagnosticList` has an impl with an `into_result` implementation (that returns `Result<(), Self>`, but you can probably have `Result<(), FederationError>).

duckki:
@SimonSapin @lrlna Oh, Simon's right. `FieldSet.parse_and_validate` is indeed specific for field set arguments. So, it makes sense to check aliases/directives in the `validate_field_set` function. It took me a long time to get here, but I want to get this right. I'll move the check into `validate_field_set` and refactor this PR as necessary.

duckki:
Based on conversation on [Slack](https://apollograph.slack.com/archives/C05263RUETS/p1712852061251609?thread_ts=1712836923.273549&cid=C05263RUETS), we are leaving the alias check in this repo, not moving it into apollo-rs repo.

lrlna (approved):
thank you thank you!

PR #255, opened 2024-04-09
add Renée and Simon to codeowners

lrlna:


goto-bus-stop (approved):

dariuszkuc (approved):
We should eventually update it to the whole GH teams (instead of individuals)

PR #257, opened 2024-04-12
Expose subgraphs schemas to crate consumers

SimonSapin:
This is to be used in Router integration


goto-bus-stop (approved):
We have `Supergraph::extract_subgraphs()` but that'd be doing a bunch of duplicate work I guess 🤔  The Supergraph type's API needs to be cleaned up anyways, makes sense to use the QueryPlanner object since you'll already have a reference to it

PR #259, opened 2024-04-12
tests: port extractSubgraphFromSupergraph tests

lennyburdette:
addresses https://apollographql.atlassian.net/browse/FED-155

Some tests were specific to bugs or issues in Federation 1.0 supergraphs, so I didn't port them


src/query_graph/extract_subgraphs_from_supergraph.rs
```diff
@@ -2067,3 +2067,769 @@ fn maybe_dump_subgraph_schema(subgraph: FederationSubgraph, message: &mut String
...
+                        sdl: |
+                            type D {
+                                d: String
+                            }
+
+            pbpaste | rover supergraph compose --config - | pbcopy
+
+        * The general idea being that all types are "value types" so ends with no 'join__type' in the supergraph (so
+        * `extractSubgraphsFromSupergraph` will have to initially assume that all types are in all subgraphs), but
+        * due to the `@provides`, `B.b` has a `join__field`, which makes it be only in "ServiceA". As as result,
```
dariuszkuc, line 2126: I thought `@provides` is only applicable on the entities? Was that supergraph v1 quirk?

lennyburdette: I honestly didn't read this comment completely — this is absolutely a supergraph v1 note. "will have to initially assume that all types are in all subgraphs" isn't true in v2 when all types are annotated with `join__type` (you can see how v1 supergraphs [leave off annotations here](https://github.com/apollographql/federation/blob/main/internals-js/src/__tests__/extractSubgraphsFromSupergraph.test.ts#L57-L75).)

dariuszkuc: we are not planning to support v1 supergraphs so unsure whether we should be porting those tests over?

lennyburdette: i ported over the tests I could convert to fed2. i can change this note to make that clearer. 

i tried converting all of them but:
* one test always failed — the test was asserting that extraction was doing the "wrong" thing
* two tests won't compose in fed 2

lennyburdette: but after going through this process, if you decided these tests are worth porting, i would not be offended

lrlna: I removed this test, but will keep the other ones.

lrlna (approved):

PR #262, opened 2024-04-18
implement NormalizedSelection::has_defer()

lrlna:
Along with `NormalizedSelection::has_defer()`, additionally implements:
* `NormalizedFieldSelection::has_defer()`
* `NormalizedFragment::has_defer()` (for `NormalizedOperation::without_defer()` implementation)
* `NormalizedFragmentSpreadSelection::has_defer()`
* `NormalizedInlineFragmentSelection::has_defer()`
* `NormalizedSelectionSet::has_defer` (for `NormalizedOperation::without_defer()` implementation)

### DIFF
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -573,33 +573,12 @@ impl NormalizedSelection {
 
     pub(crate) fn has_defer(&self) -> bool {
         match self {
-            NormalizedSelection::Field(field_selection) => {
-                // @defer cannot be on a field at the moment
-                field_selection.field.data().directives.has("defer")
-                    || field_selection
-                        .selection_set
-                        .as_ref()
-                        .is_some_and(|s| s.has_defer())
-            }
+            NormalizedSelection::Field(field_selection) => field_selection.has_defer(),
             NormalizedSelection::FragmentSpread(fragment_spread_selection) => {
-                fragment_spread_selection
-                    .spread
-                    .data
-                    .directives
-                    .has("defer")
-                    || fragment_spread_selection.selection_set.has_defer()
+                fragment_spread_selection.has_defer()
             }
             NormalizedSelection::InlineFragment(inline_fragment_selection) => {
-                inline_fragment_selection
-                    .inline_fragment
-                    .data()
-                    .directives
-                    .has("defer")
-                    || inline_fragment_selection
-                        .selection_set
-                        .selections
-                        .values()
-                        .any(|s| s.has_defer())
+                inline_fragment_selection.has_defer()
             }
         }
     }
@@ -1031,6 +1010,10 @@ impl NormalizedFragmentSpreadSelection {
                 selection_set: named_fragment.selection_set.clone(),
             },
         ))))
+    }
+
+    pub(crate) fn has_defer(&self) -> bool {
+        self.spread.data.directives.has("defer") || self.selection_set.has_defer()
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/query_plan/operation.rs
```diff
@@ -601,8 +571,19 @@ impl NormalizedSelection {
         }
     }
 
-    pub(crate) fn has_defer(&self) -> Result<bool, FederationError> {
-        todo!()
+    pub(crate) fn has_defer(&self) -> bool {
```
lrlna, line 574: I am not sure if the return type of the stub - `Result<bool, FederationError>` - was intentional. The JS implementation just returns a bool, and I am not sure in which cases we'd want to throw an error. Let me know if the `Result` was there for a reason.

lrlna: @SimonSapin iirc this was your original stub. did you have a plan for the `Result`?

dariuszkuc: I don't see a case where this could potentially throw. Lets keep it as `bool` and change it later on if needed.

dariuszkuc: I believe attempting to specify `@defer` on a field selection could result in an error? Since parsed document is validated I guess maybe this was a safeguard so we don't transform selection set to something invalid? I guess if you incorrectly try to normalize inline fragment spread to lift up selection you could end up with something invalid.... but that would be a bug in the code?

dariuszkuc (approved):

src/query_plan/operation.rs
```diff
@@ -759,6 +744,13 @@ pub(crate) mod normalized_field_selection {
         pub(crate) selection_set: Option<NormalizedSelectionSet>,
     }
 
+    impl NormalizedFieldSelection {
```
dariuszkuc, line 747: can we move all those selections `has_defer`s to top level impls? per previous review comments the inner mods created to keep the key/value in sync but all the other stuff should be moved outside

src/query_plan/operation.rs
```diff
@@ -857,6 +849,12 @@ pub(crate) mod normalized_fragment_spread_selection {
         pub(crate) selection_set: NormalizedSelectionSet,
     }
 
+    impl NormalizedFragmentSpreadSelection {
+        pub(crate) fn has_defer(&self) -> bool {
+            self.spread.data.directives.has("defer")
```
dariuszkuc, line 854: we do store the selection set on the spread selection now so we should also whether selection set has defer

src/query_plan/operation.rs
dariuszkuc: JS code doesn't throw so we should keep it as `bool`

dariuszkuc (changes requested):
can we move all those selections `has_defer`s to top level impls? per previous review comments the inner mods created to keep the key/value in sync but all the other stuff should be moved outside

src/query_plan/operation.rs
lrlna: good catch, ty!

goto-bus-stop:
This PR adds the `has_defer()` impl blocks inside the nested modules inside `operation.rs`, which are meant to only contain the functions that need to maintain type invariants (like selection keys and selections staying in sync). `has_defer()` doesn't mutate and doesn't require keeping things in sync so it can be at the "root" of `operation.rs`.

Currently the module separation doesn't actually do anything as the fields are `pub(crate)` anyways. if we want to enforce invariants the fields should be private

lrlna:
@goto-bus-stop ah, I see, thanks for clarifying. Removed them from the internal mod then.

dariuszkuc (approved):
Changes look fine but I kind of preferred the impls per selection vs just putting all the logic into the single method-> we do have top level selection impls (outside of mods) so my comment was about just moving the functions around from inner mod to top level `operation.rs` impls

PR #264, opened 2024-04-18
Port of selection_set_is_fully_local_from_all_nodes

TylerBloom:
This PR addresses FED-42 and ports [selectionIsFullyLocalFromAllVertices](https://github.com/apollographql/federation/blob/main/query-planner-js/src/buildPlan.ts#L251).

### DIFF
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -24,8 +24,6 @@ use crate::query_plan::FetchDataRewrite;
 use crate::schema::definitions::is_composite_type;
 use crate::schema::definitions::types_can_be_merged;
 use crate::schema::definitions::AbstractType;
-use crate::schema::position::FieldDefinitionPosition;
-use crate::schema::position::UnionTypeDefinitionPosition;
 use crate::schema::position::{
     CompositeTypeDefinitionPosition, InterfaceTypeDefinitionPosition, ObjectTypeDefinitionPosition,
     SchemaRootDefinitionKind,
@@ -854,10 +852,7 @@ pub(crate) mod normalized_field_selection {
         NormalizedSelectionSet,
     };
     use crate::query_plan::FetchDataPathElement;
-    use crate::schema::position::{
-        CompositeTypeDefinitionPosition, FieldDefinitionPosition, InterfaceTypeDefinitionPosition,
-        ObjectTypeDefinitionPosition, TypeDefinitionPosition, UnionTypeDefinitionPosition,
-    };
+    use crate::schema::position::{FieldDefinitionPosition, TypeDefinitionPosition};
     use crate::schema::ValidFederationSchema;
     use apollo_compiler::ast::{Argument, Directive, DirectiveList, Name};
     use apollo_compiler::Node;
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -3478,7 +3478,9 @@ impl NormalizedInlineFragmentSelection {
         parent_type: &CompositeTypeDefinitionPosition,
         schema: &ValidFederationSchema,
     ) -> bool {
-        if &self.inline_fragment.data().parent_type_position == parent_type {
+        if &self.inline_fragment.data().parent_type_position == parent_type
+            && self.inline_fragment.data().schema == *schema
+        {
             return true;
         }
         let Some(ty) = self
@@ -3488,7 +3490,7 @@ impl NormalizedInlineFragmentSelection {
         else {
             return false;
         };
-        if self.inline_fragment.data().parent_type_position != ty {
+        if self.selection_set.type_position != ty {
             for sel in self.selection_set.selections.values() {
                 if !sel.can_add_to(&ty, schema) {
                     return false;
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -1109,41 +1109,6 @@ pub(crate) mod normalized_fragment_spread_selection {
         pub(crate) fn data(&self) -> &NormalizedFragmentSpreadData {
             &self.data
         }
-
-        pub(crate) fn can_add_to(
-            &self,
-            parent_type: &CompositeTypeDefinitionPosition,
-            schema: &ValidFederationSchema,
-        ) -> Result<bool, FederationError> {
-            if parent_type == &self.data().type_condition_position {
-                return Ok(true);
-            }
-
-            let Some(ty) = self.data().type_if_added_to(parent_type, schema)? else {
-                return Ok(false);
-            };
-
-            todo!()
-
-            /*
-            if let Some(set) = &self.data().selection_set {
-                if set.type_position != ty {
-                    if ty.is_object_type() || ty.is_interface_type() {
-                        return Err(FederationError::internal(format!(
-                            "{} should have a selection set as it's type is not a composite",
-                            self.field,
-                        )));
-                    }
-                    for sel in set.selections.values() {
-                        if !sel.can_add_to(parent_type, schema)? {
-                            return Ok(false);
-                        }
-                    }
-                }
-            }
-            Ok(true)
-            */
-        }
     }
 
     impl HasNormalizedSelectionKey for NormalizedFragmentSpread {
@@ -1225,6 +1190,43 @@ pub(crate) mod normalized_fragment_spread_selection {
                 }
             }
         }
+    }
+}
+
+impl NormalizedFragmentSpread {
+    pub(crate) fn can_add_to(
+        &self,
+        parent_type: &CompositeTypeDefinitionPosition,
+        schema: &ValidFederationSchema,
+    ) -> Result<bool, FederationError> {
+        if parent_type == &self.data().type_condition_position {
+            return Ok(true);
+        }
+
+        let Some(ty) = self.data().type_if_added_to(parent_type, schema)? else {
+            return Ok(false);
+        };
+
+        todo!()
+
+        /*
+        if let Some(set) = &self.data().selection_set {
+            if set.type_position != ty {
+                if ty.is_object_type() || ty.is_interface_type() {
+                    return Err(FederationError::internal(format!(
+                        "{} should have a selection set as it's type is not a composite",
+                        self.field,
+                    )));
+                }
+                for sel in set.selections.values() {
+                    if !sel.can_add_to(parent_type, schema)? {
+                        return Ok(false);
+                    }
+                }
+            }
+        }
+        Ok(true)
+        */
     }
 }
 
@@ -1483,7 +1485,7 @@ pub(crate) mod normalized_inline_fragment_selection {
     /// graph paths.
     #[derive(Debug, Clone, PartialEq, Eq, Hash)]
     pub(crate) struct NormalizedInlineFragment {
-        data: NormalizedInlineFragmentData,
+        pub(super) data: NormalizedInlineFragmentData,
         key: NormalizedSelectionKey,
     }
 
@@ -1506,14 +1508,6 @@ pub(crate) mod normalized_inline_fragment_selection {
             let mut data = self.data().clone();
             data.type_condition_position = new;
             Self::new(data)
-        }
-
-        pub(crate) fn can_add_to(
-            &self,
-            parent_type: &CompositeTypeDefinitionPosition,
-            schema: &ValidFederationSchema,
-        ) -> Result<bool, FederationError> {
-            self.data.can_add_to(parent_type, schema)
         }
         pub(crate) fn with_updated_directives(
             &self,
@@ -3035,20 +3029,22 @@ impl NormalizedFieldSelection {
         parent_type: &CompositeTypeDefinitionPosition,
         schema: &ValidFederationSchema,
     ) -> Result<bool, FederationError> {
-        match (parent_type, &self.field.data().field_position) {
-            (
-                CompositeTypeDefinitionPosition::Object(ty_a),
-                FieldDefinitionPosition::Object(ty_b),
-            ) if ty_a.type_name == ty_b.type_name => return Ok(true),
-            (
-                CompositeTypeDefinitionPosition::Interface(ty_a),
-                FieldDefinitionPosition::Interface(ty_b),
-            ) if ty_a.type_name == ty_b.type_name => return Ok(true),
-            (
-                CompositeTypeDefinitionPosition::Union(ty_a),
-                FieldDefinitionPosition::Union(ty_b),
-            ) if ty_a.type_name == ty_b.type_name => return Ok(true),
-            _ => {}
+        if &self.field.data().schema == schema {
+            match (parent_type, &self.field.data().field_position) {
+                (
+                    CompositeTypeDefinitionPosition::Object(ty_a),
+                    FieldDefinitionPosition::Object(ty_b),
+                ) if ty_a.type_name == ty_b.type_name => return Ok(true),
+                (
+                    CompositeTypeDefinitionPosition::Interface(ty_a),
+                    FieldDefinitionPosition::Interface(ty_b),
+                ) if ty_a.type_name == ty_b.type_name => return Ok(true),
+                (
+                    CompositeTypeDefinitionPosition::Union(ty_a),
+                    FieldDefinitionPosition::Union(ty_b),
+                ) if ty_a.type_name == ty_b.type_name => return Ok(true),
+                _ => {}
+            }
         }
 
         let Some(ty) = self.field.type_if_added_to(parent_type, schema)? else {
@@ -3057,12 +3053,6 @@ impl NormalizedFieldSelection {
 
         if let Some(set) = &self.selection_set {
             if set.type_position != ty {
-                if ty.is_object_type() || ty.is_interface_type() {
-                    return Err(FederationError::internal(format!(
-                        "{} should have a selection set as it's type is not a composite",
-                        self.field,
-                    )));
-                }
                 for sel in set.selections.values() {
                     if !sel.can_add_to(parent_type, schema)? {
                         return Ok(false);
@@ -3723,6 +3713,14 @@ impl NormalizedInlineFragment {
             // can_rebase = false, condition = undefined
             (false, None)
         }
+    }
+
+    pub(crate) fn can_add_to(
+        &self,
+        parent_type: &CompositeTypeDefinitionPosition,
+        schema: &ValidFederationSchema,
+    ) -> Result<bool, FederationError> {
+        self.data.can_add_to(parent_type, schema)
     }
 }
 
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -517,12 +517,6 @@ pub(crate) trait HasNormalizedSelectionKey {
     fn key(&self) -> NormalizedSelectionKey;
 }
 
-impl HasNormalizedSelectionKey for NormalizedSelectionKey {
-    fn key(&self) -> NormalizedSelectionKey {
-        self.clone()
-    }
-}
-
 /// An analogue of the apollo-compiler type `Selection` that stores our other selection analogues
 /// instead of the apollo-compiler types.
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -3084,7 +3078,7 @@ impl NormalizedField {
         schema: &ValidFederationSchema,
     ) -> Option<CompositeTypeDefinitionPosition> {
         let data = self.data();
-        if &data.field_position.parent() == parent_type && &data.schema == schema {
+        if data.field_position.parent() == *parent_type && data.schema == *schema {
             let base_ty_name = data
                 .field_position
                 .get(schema.schema())
--- a/src/query_plan/operation.rs
+++ b/src/query_plan/operation.rs
@@ -24,6 +24,7 @@ use crate::query_plan::FetchDataRewrite;
 use crate::schema::definitions::is_composite_type;
 use crate::schema::definitions::types_can_be_merged;
 use crate::schema::definitions::AbstractType;
+use crate::schema::position::FieldDefinitionPosition;
 use crate::schema::position::{
     CompositeTypeDefinitionPosition, InterfaceTypeDefinitionPosition, ObjectTypeDefinitionPosition,
     SchemaRootDefinitionKind,
@@ -517,6 +518,12 @@ pub(crate) trait HasNormalizedSelectionKey {
     fn key(&self) -> NormalizedSelectionKey;
 }
 
+impl HasNormalizedSelectionKey for NormalizedSelectionKey {
+    fn key(&self) -> NormalizedSelectionKey {
+        self.clone()
+    }
+}
+
 /// An analogue of the apollo-compiler type `Selection` that stores our other selection analogues
 /// instead of the apollo-compiler types.
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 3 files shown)


src/query_plan/operation.rs
```diff
@@ -782,6 +808,49 @@ pub(crate) mod normalized_field_selection {
...
+                return Ok(Some(digest));
+            }
+            if data.name() == parent_type.type_name() {
+                todo!()
+            }
+            /*
+            if (this.name === typenameFieldName) {
+                return parentType.typenameField()?.type;
+            }
+            */
```
TylerBloom, line 841: Remaining question: What is the RS equivalent of `typenameField`?

src/query_plan/operation.rs
```diff
@@ -782,6 +808,49 @@ pub(crate) mod normalized_field_selection {
...
+                // parent_type.field(self.data.name())?
+                todo!()
+            } else {
+                Ok(None)
+            }
+            /*
+            return this.canRebaseOn(parentType)
+                ? parentType.field(this.name)?.type
+                : undefined;
+            */
```
TylerBloom, line 852: Remaining question: We can the a `FieldDefinitionPosition`. Is it sound to construct a `CompositeTypeDefinitionPosition` with it (i.e. map the variants and remove the `field_name` from `FieldDefinitionPosition`?

src/query_plan/operation.rs
```diff
@@ -1124,6 +1281,40 @@ pub(crate) mod normalized_inline_fragment_selection {
                 Ok(None)
             }
         }
+
+        pub(crate) fn can_add_to(
+            &self,
+            parent_type: &CompositeTypeDefinitionPosition,
+            schema: &ValidFederationSchema,
+        ) -> Result<bool, FederationError> {
```
TylerBloom, line 1289: Remaining question: I'm not sure how to port this from JS. Inheritance made this hard to parse out. A pointer on where to look would be helpful.

src/query_plan/operation.rs
```diff
@@ -873,6 +943,41 @@ pub(crate) mod normalized_fragment_spread_selection {
         pub(crate) fn data(&self) -> &NormalizedFragmentSpreadData {
             &self.data
         }
+
+        pub(crate) fn can_add_to(
+            &self,
+            parent_type: &CompositeTypeDefinitionPosition,
+            schema: &ValidFederationSchema,
+        ) -> Result<bool, FederationError> {
```
TylerBloom, line 951: Remaining question: I'm not sure how to port this from JS. Inheritance made this hard to parse out. A pointer on where to look would be helpful.

src/query_plan/operation.rs
```diff
@@ -897,6 +1002,50 @@ pub(crate) mod normalized_fragment_spread_selection {
         pub(crate) selection_id: SelectionId,
     }
 
+    impl NormalizedFragmentSpreadData {
+        pub(crate) fn type_if_added_to(
+            &self,
+            parent_type: &CompositeTypeDefinitionPosition,
+            schema: &ValidFederationSchema,
+        ) -> Result<Option<CompositeTypeDefinitionPosition>, FederationError> {
```
TylerBloom, line 1010: Remaining question: I'm not sure how to port this from JS. Inheritance made this hard to parse out. A pointer on where to look would be helpful.

src/query_plan/operation.rs
lrlna: I think you can use the `parent_type.introspection_typename_field()` to get the `__typename`

src/query_plan/query_planning_traversal.rs
```diff
@@ -427,10 +429,58 @@ impl<'a> QueryPlanningTraversal<'a> {
 
     fn selection_set_is_fully_local_from_all_nodes(
         &self,
-        _selection: &NormalizedSelectionSet,
-        _nodes: &IndexSet<NodeIndex>,
+        selection: &NormalizedSelectionSet,
+        nodes: &IndexSet<NodeIndex>,
     ) -> Result<bool, FederationError> {
-        todo!()
+        // To guarantee that the selection is fully local from the provided vertex/type, we must have:
```
dariuszkuc, line 435: looks like we are missing this [block](https://github.com/apollographql/federation/blob/main/query-planner-js/src/buildPlan.ts#L257-L265) which checks for inconsistent abstract types on fragment conditions?

src/query_plan/operation.rs
```diff
@@ -2776,6 +2965,49 @@ impl NormalizedFieldSelection {
     pub(crate) fn has_defer(&self) -> bool {
         self.field.has_defer() || self.selection_set.as_ref().is_some_and(|s| s.has_defer())
     }
+
+    fn can_add_to(
+        &self,
+        parent_type: &CompositeTypeDefinitionPosition,
+        schema: &ValidFederationSchema,
+    ) -> Result<bool, FederationError> {
+        match (parent_type, &self.field.data().field_position) {
```
dariuszkuc, line 2974: JS code actually checks whether we are pointing at the same object -> here we need to check the type names AND whether we point to the same schema (I did the same check in rebasing logic [here](https://github.com/apollographql/federation-next/blob/main/src/query_plan/operation.rs#L2717-L2722))

src/query_plan/operation.rs
```diff
@@ -2776,6 +2965,49 @@ impl NormalizedFieldSelection {
...
+            ) if ty_a.type_name == ty_b.type_name => return Ok(true),
+            _ => {}
+        }
+
+        let Some(ty) = self.field.type_if_added_to(parent_type, schema)? else {
+            return Ok(false);
+        };
+
+        if let Some(set) = &self.selection_set {
+            if set.type_position != ty {
```
dariuszkuc, line 2995: do we need to unwrap the `ty`? i.e. the `baseType` from JS side of things - I believe there is already a corresponding method for this in `apollo-rs`

src/query_plan/operation.rs
```diff
@@ -2776,6 +2965,49 @@ impl NormalizedFieldSelection {
...
+                if ty.is_object_type() || ty.is_interface_type() {
```
dariuszkuc, line 2996: I think this should should also check for union?

src/query_plan/operation.rs
dariuszkuc: You should add those functions to top level impls (outside of inner mods)

* `NormalizedFieldData` (which would be invoked from field selection) should implement `type_if_added_to` -> [JS field `typeIfAddedTo`](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L349-L362)
* On JS side of things `FragmentElement` is used to represent both inline fragment data and spread data so it should be the same logic for both.... but since `can_add_to` is always true for spreads, we just need to implement `casted_type_if_added_to` for `NormalizedInlineFragmentData` (as it would be invoked from its `can_add_to`) -> [JS inline fragment `castedTypeIfAddedTo`](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L524-L531)

src/query_plan/operation.rs
dariuszkuc: As per my other comment -> `can_add_to` methods should be on the corresponding top level (outside of inner mod) `XyzSelection` impls. 

* `NormalizedFragmentSpreadSelection` ->[ should always return `true`](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L3689-L3693)
* `InlineFragmentSelection` -> [checks if casted type can be added and then validates the selection set](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L3330-L3344)
* `FieldSelection` -> you already got it implemented ([JS logic](https://github.com/apollographql/federation/blob/main/internals-js/src/operations.ts#L3091-L3107)

dariuszkuc (changes requested):

src/query_plan/query_planning_traversal.rs
```diff
@@ -427,10 +429,58 @@ impl<'a> QueryPlanningTraversal<'a> {
 
     fn selection_set_is_fully_local_from_all_nodes(
         &self,
-        _selection: &NormalizedSelectionSet,
-        _nodes: &IndexSet<NodeIndex>,
+        selection: &NormalizedSelectionSet,
+        nodes: &IndexSet<NodeIndex>,
     ) -> Result<bool, FederationError> {
```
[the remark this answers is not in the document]
TylerBloom: The call to `can_rebase_on` can return an error.
```rust
            if n.has_reachable_cross_subgraph_edges || selection.can_rebase_on(&parent_ty)? {
                return Ok(false);
            }
```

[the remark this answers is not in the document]
dariuszkuc: yeah but the `can_rebase_on` probably should return just `bool` as well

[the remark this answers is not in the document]
TylerBloom: That is probably true. Once the other `can_add_to` methods are implemented, I'll check if we can just return a `bool`. One change that I'll push with the updates from the comments, is this on L462
```rust
            // Change this
            let n = &qg[*node];
            // To this:
            let n = self.parameters.federated_query_graph.node_weight(*node)?;
```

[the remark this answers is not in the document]
TylerBloom: That index operation *shouldn't* panic, but "should" can be a dangerous word.

src/query_plan/operation.rs
TylerBloom: Hmm.. after looking at this again, I don't think this check is needed. The [original line](https://github.com/apollographql/federation/blob/bbf83946800d0e6f83307e1941cd7052f8a3c38d/internals-js/src/operations.ts#L3103) checks for composite types (which I got my wired crossed with). However, `ty` is an `CompositeTypeDefinitionPosition` and can only be a compose type.

src/query_plan/operation.rs
TylerBloom: The JS code does call `baseType`, but I don't think it is needed here. The equivalent is `Type::inner_named_type`, which return the `NamedType` (i.e. a `Name`) san any markers of being nullable or in a list. Here, `ty` is a composition type position, so that extra data should be present.

src/query_plan/query_planning_traversal.rs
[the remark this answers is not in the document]
TylerBloom: After some cleanup, this line is the only one that can error.

src/query_plan/operation.rs
```diff
@@ -869,6 +891,34 @@ pub(crate) mod normalized_field_selection {
...
+                            InterfaceTypeDefinitionPosition::new(ty.type_name.clone()),
+                        )
+                    }
+                    FieldDefinitionPosition::Union(ty) => CompositeTypeDefinitionPosition::Union(
+                        UnionTypeDefinitionPosition::new(ty.type_name.clone()),
+                    ),
+                };
+                return Some(digest);
+            }
+            if data.name() == parent_type.type_name() {
```
dariuszkuc, line 916: this actually should be checking whether field name is [`__typename`](https://github.com/apollographql/federation-next/blob/main/src/query_plan/operation.rs#L51)

src/query_plan/operation.rs
```diff
@@ -869,6 +891,34 @@ pub(crate) mod normalized_field_selection {
             &mut self.data.sibling_typename
         }
 
+        pub(crate) fn type_if_added_to(
+            &self,
+            parent_type: &CompositeTypeDefinitionPosition,
+            schema: &ValidFederationSchema,
+        ) -> Option<CompositeTypeDefinitionPosition> {
```
dariuszkuc, line 898: based on the JS implementation this should be returning actual schema type (i.e. `apollo-compiler` enum `Type`)

src/query_plan/operation.rs
```diff
@@ -869,6 +891,34 @@ pub(crate) mod normalized_field_selection {
...
+                        UnionTypeDefinitionPosition::new(ty.type_name.clone()),
+                    ),
+                };
+                return Some(digest);
+            }
+            if data.name() == parent_type.type_name() {
+                return Some(parent_type.introspection_typename_field().parent());
+            }
+            self.can_rebase_on(parent_type, schema)
+                .then(|| data.field_position.parent())
```
dariuszkuc, line 920: this should be returning actual field type -> note that field parent is different than the parent (first check should be checking whether field is on the same parent equivalence), e.g. it can be from different schema so this should be doing something along those lines

```rs
// get parent field -> get field from schema -> get type
parent_type.field(data.field_position.field_name())?.get(schema.schema())?.ty
```

src/query_plan/operation.rs
```diff
@@ -869,6 +891,34 @@ pub(crate) mod normalized_field_selection {
             &mut self.data.sibling_typename
         }
 
+        pub(crate) fn type_if_added_to(
```
dariuszkuc, line 894: can you move it to the top level `NormalizedField` impl?

src/query_plan/operation.rs
```diff
@@ -869,6 +891,34 @@ pub(crate) mod normalized_field_selection {
...
+            &self,
+            parent_type: &CompositeTypeDefinitionPosition,
+            schema: &ValidFederationSchema,
+        ) -> Option<CompositeTypeDefinitionPosition> {
+            let data = self.data();
+            if &data.field_position == parent_type {
```
dariuszkuc, line 900: this should be returning something like
```rs
if data.field_position.parent() == *parent_type && data.schema == *schema {
  let field = data.field_position.get(schema.schema())?;
  return Ok(Some(field.ty.clone()));
}
```

src/query_plan/operation.rs
```diff
@@ -1294,7 +1344,7 @@ pub(crate) mod normalized_inline_fragment_selection {
     /// graph paths.
     #[derive(Debug, Clone, PartialEq, Eq, Hash)]
     pub(crate) struct NormalizedInlineFragment {
-        data: NormalizedInlineFragmentData,
+        pub(super) data: NormalizedInlineFragmentData,
```
dariuszkuc, line 1347: you can access it through a `data()` getter -> this is private on purpose

src/query_plan/operation.rs
```diff
@@ -1371,6 +1420,48 @@ pub(crate) mod normalized_inline_fragment_selection {
                 Ok(None)
             }
         }
+
+        pub(super) fn casted_type_if_add_to(
+            &self,
+            parent_type: &CompositeTypeDefinitionPosition,
+        ) -> Option<CompositeTypeDefinitionPosition> {
+            if &self.parent_type_position == parent_type {
```
dariuszkuc, line 1428: you need to check whether the types are on the same schema

src/query_plan/operation.rs
```diff
@@ -2776,6 +2873,44 @@ impl NormalizedFieldSelection {
...
         self.field.has_defer() || self.selection_set.as_ref().is_some_and(|s| s.has_defer())
     }
+
+    fn can_add_to(
+        &self,
+        parent_type: &CompositeTypeDefinitionPosition,
+        schema: &ValidFederationSchema,
+    ) -> bool {
+        if &self.field.data().schema == schema {
+            match (parent_type, &self.field.data().field_position) {
```
dariuszkuc, line 2883: I believe you should be matching field parent against the parent (and not field type vs parent type) -> i.e. 

``` &self.field.data().schema == schema && &self.field.data().field_position.parent() == parent_type```

File: src/query_plan/operation.rs (hunk 12 above)
Comment (dariuszkuc) [reply src/query_plan/operation.rs:898]: actually since the only place calling into this method is then unwrapping the type (`baseType`) we could probably convert the GraphQL type to a `CompositeTypeDefinitionPosition`

**Review by dariuszkuc — DISMISSED** [review]
I think it would be a good idea to add some unit tests for `selection_set_is_fully_local_from_all_nodes`

File: src/query_plan/operation.rs (hunk 12 above)
Comment (TylerBloom) [reply src/query_plan/operation.rs:898]: > actually since the only place calling into this method is then unwrapping the type (baseType) we could probably convert the GraphQL type to a CompositeTypeDefinitionPosition

This was my thought. In the JS code, they ensure that the object is a composite type. This also avoids any (currently) unnecessary `TryInto` calls. Either way, there should be a port note explaining this.

File: src/query_plan/operation.rs (hunk 11 above)
Comment (TylerBloom) [reply src/query_plan/operation.rs:916]: I have the if condition changed to check for `__typename`. Do you also want to the `return` call changed or is it ok as is?

File: src/query_plan/operation.rs (hunk 15 above)
Comment (TylerBloom) [reply src/query_plan/operation.rs:900]: `field.ty` is an AST `Type`, not a composite type. Knowing where this method is called, we could try looking up the `ExtenedType` and then converting that into a composite type, and flatten everything using `Result::ok` (if this method would return anything other than `Some`/`Ok(Some())`, the calling method will return `false`, so dropping the `Err` doesn't not cause any info loss). This would look (roughly) like this:
```rust
if &data.field_position.parent() == parent_type && &data.schema == schema {
  let base_ty_name = data.field_position.get(schema.schema()).ok()?.ty.inner_named_type();
   return schema.get_type(base_ty_name.clone()).and_then(CompositeTypeDefinitionPosition::try_from).ok();
}
```

src/query_plan/query_planning_traversal.rs
```diff
@@ -432,10 +434,50 @@ impl<'a> QueryPlanningTraversal<'a> {
...
+        if !nodes.is_empty()
+            && selection.selections.keys().any(|k| match k {
+                NormalizedSelectionKey::InlineFragment {
+                    type_condition: Some(name),
+                    ..
+                } => self
+                    .parameters
+                    .abstract_types_with_inconsistent_runtime_types
+                    .iter()
+                    .any(|ty| ty.type_name() == name),
```
goto-bus-stop, line 459: hm, I think this is the right thing to do right now, but I don't love that we aren't using the Set-ness of `abstract_types_with_inconsistent_runtime_types`. maybe its type needs to be less strict, so we can use `.contains()` even when we don't have an `AbstractTypeDefinitionPosition` like here.

src/schema/position.rs
```diff
@@ -582,6 +582,26 @@ pub(crate) enum FieldDefinitionPosition {
     Union(UnionTypenameFieldDefinitionPosition),
 }
 
+impl PartialEq<CompositeTypeDefinitionPosition> for FieldDefinitionPosition {
```
goto-bus-stop, line 585: This should not be an eq implementation in my opinion. This means that `A.f == A`, which is not what I'd ever expect when reading an `==`. I think we can replace uses of this impl by `field_definition_position.parent() == composite_type_position` instead?

src/query_plan/operation.rs
```diff
@@ -517,6 +517,12 @@ pub(crate) trait HasNormalizedSelectionKey {
     fn key(&self) -> NormalizedSelectionKey;
 }
 
+impl HasNormalizedSelectionKey for NormalizedSelectionKey {
+    fn key(&self) -> NormalizedSelectionKey {
+        self.clone()
+    }
+}
```
goto-bus-stop, line 524: This looks a bit smelly. Why is it necessary?

src/query_plan/operation.rs
```diff
@@ -2985,6 +3077,31 @@ impl NormalizedField {
...
         false
     }
+
+    pub(crate) fn type_if_added_to(
+        &self,
+        parent_type: &CompositeTypeDefinitionPosition,
+        schema: &ValidFederationSchema,
+    ) -> Option<CompositeTypeDefinitionPosition> {
+        let data = self.data();
+        if &data.field_position.parent() == parent_type && &data.schema == schema {
```
goto-bus-stop, line 3087: stylistic nit but writing the references the other way around makes this a bit easier to read IMO, so you avoid the `&& &`. It functions the same, since `==` operates on a reference it doesn't actually need to move out of the reference when you write `*parent_type`.
```suggestion
        if data.field_position.parent() == *parent_type && data.schema == *schema {
```

src/query_plan/query_planning_traversal.rs
```diff
@@ -432,10 +434,50 @@ impl<'a> QueryPlanningTraversal<'a> {
...
+        // - the type must be compositeType (mostly just ensuring the selection make sense).
+        // - everything in the selection must be avaiable in the type (which `rebaseOn` essentially validates).
+        // - the selection must not "type-cast" into any abstract type that has inconsistent runtimes acrosse subgraphs. The reason for the
+        //   later condition is that `selection` is originally a supergraph selection, but that we're looking to apply "as-is" to a subgraph.
+        //   But suppose it has a `... on I` where `I` is an interface. Then it's possible that `I` includes "more" types in the supergraph
+        //   than in the subgraph, and so we might have to type-explode it. If so, we cannot use the selection "as-is".
+        // PORT_NOTE: The JS code performs the last check lazily. Instead of that, this check is
+        // skipped if `nodes` is empty.
+        if !nodes.is_empty()
+            && selection.selections.keys().any(|k| match k {
```
goto-bus-stop, line 451: Actually, can we match on the *values* instead of the keys here? I think we should treat keys as opaque if we can. Maybe they should not even be exposed from the operation module.

goto-bus-stop (changes requested):
I looked over most of the methods and I have a few more comments. I'll look over the remainder after standup. The amount of comments doesn't really reflect it but I actually am happy with what I see so far :)

src/query_plan/operation.rs
TylerBloom: Probably should be removed. At one point, I thought I was going to need to use this, but I ended up not needing it.

src/schema/position.rs
TylerBloom: Like the comment above, this was something that I thought I was going to need but didn't (I was trying to find my way through the maze of the codebase and got a bit lost). Sorry, I should have caught these. Unfortunately, the dead code lint doesn't catch unused trait impls (even for fully private types) :cry:

goto-bus-stop: great, no worries! 😄

goto-bus-stop (changes requested):

src/query_plan/operation.rs
```diff
@@ -2985,6 +3071,31 @@ impl NormalizedField {
...
+            return schema
+                .get_type(base_ty_name.clone())
+                .and_then(CompositeTypeDefinitionPosition::try_from)
+                .ok();
+        }
+        if data.name() == &TYPENAME_FIELD {
+            return Some(parent_type.introspection_typename_field().parent());
+        }
+        self.can_rebase_on(parent_type, schema)
+            .then(|| data.field_position.parent())
```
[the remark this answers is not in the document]
TylerBloom: Gotcha, so I would need to replace
```rust
            return Some(parent_type.introspection_typename_field().parent());
```
with something like
```rust
            let type_name = parent_type.introspection_typename_field()
                            .get(schema.schema())
                            .ok()?
                            .ty
                            .inner_named_type();
            return schema.try_get_type(type_name.clone())?.try_into().ok();
```
For the other issue, this
```rust
        self.can_rebase_on(parent_type, schema)
            .then(|| data.field_position.parent())
```
would be replaced with
```rust
        // Implement From<FieldPositionDefinition> for CompositeTypeDefinition by dropping `field_name` from all variants.
        self.can_rebase_on(parent_type, schema)
            .then(|| data.field_position.into())
```
Are these solution (more) correct?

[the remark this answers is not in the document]
goto-bus-stop: `data.field_position.into()` would actually be doing the same thing as `.parent()` in that case, so you would still be returning `Query` when it should return `ReturnType`. `data.field_position` is referring to the `Query.field` definition in the original subgraph, but we need to find the return type of the field *after rebasing* (likely onto a different subgraph). So you have to jump through the same hoops as for `__typename`:

```rust
// assuming we are considering selecting this field from a type named "Query2":
// "Query.field" -> "Query2.field"
parent_type.field(data.field_position.field_name())
// look it up in the schema we are rebasing on
  .get(schema.schema())
  ... // the rest of the dance is the same as for __typename.
```

goto-bus-stop (approved):
Okay two final suggestions. I keep overlooking stuff apparently!

PR #268, opened 2024-04-23
use a map for headers

lennyburdette:
We use a list of `{name:, value?:, as?:}` object because we can't use an object — header names aren't valid GraphQL object keys. But it really is a map, especially so we can easily merge the `@connect` headers into the `@source` headers, overriding by name. Also, the `value` field can be a list of values.


nicholascioli (approved):
LGTM!

PR #269, opened 2024-04-23
Draft WASM validation logic

nicholascioli:
This commit adds a partial implementation of validation of a string schema, which will be useful in WASM contexts where operating on Rust types does not work.

This implementation is partial as it can still trigger a panic if a directive is invalid.


src/sources/connect/wasm_validators.rs
```diff
@@ -0,0 +1,245 @@
...
+            reason: reason.to_string(),
+            error: e.into(),
+        })
+    }
+
+    // First we need to create an actul schema from the provided one
+    let combined_schema = format!(
+        "{}\n{}\n{}",
+        subgraph_schema,
+        constants::TEMP_FEDERATION_DEFINITIONS,
```
lennyburdette, line 48: this shouldn't be necessary, let me check the JS code real quick

lennyburdette: yep — the subgraphs [at this point](https://github.com/apollographql/federation/blob/main/composition-js/src/compose.ts#L57) are already preprocessed and made valid using the specs (link, federation, connect) so we don't need to edit them before parsing/validating.

nicholascioli: Sounds good! I'll fix this now

lennyburdette (approved):

src/sources/connect/wasm_validators.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn error_with_phony_loc(reason: &str, e: impl Into<FederationError>) -> Box<ValidationError> {
+        Box::new(ValidationError {
+            range: Default::default(),
+            reason: reason.to_string(),
+            error: e.into(),
+        })
+    }
+
+    // First we need to create an actul schema from the provided one
+    let schema = apollo_compiler::Schema::parse(subgraph_schema, "")
```
lennyburdette, line 45: we can use parse_and_validate i think

src/sources/connect/wasm_validators.rs
```diff
@@ -0,0 +1,89 @@
...
+    let schema = schema
+        .validate()
+        .map_err(|e| error_with_phony_loc("failed to validate schema", e))?;
+    let schema = ValidFederationSchema::new(schema)
+        .map_err(|e| error_with_phony_loc("failed to parse as federation schema", e))?;
+
+    // Process the @source directives
+    let sources = schema
+        .referencers()
+        .get_directive(&SOURCE_DIRECTIVE_NAME_IN_SPEC)
```
lennyburdette, line 58: this will error if it can't find any sources — before we do anything, we should first check to see if the `@link` is present using `schema.metadata()` (and return early if it's not), and we should ignore the Err case for sources since they're optional.

nicholascioli:
This is no longer needed, as it is out of date and will be made anew.

PR #271, opened 2024-04-24
Port QueryGraph::get_locally_satisfiable_key

goto-bus-stop:
This is used when resolving conditions. Compare:

https://github.com/apollographql/federation/blob/bbf83946800d0e6f83307e1941cd7052f8a3c38d/query-graphs-js/src/graphPath.ts#L1768-L1781

### DIFF
--- a/src/query_graph/mod.rs
+++ b/src/query_graph/mod.rs
@@ -2,6 +2,7 @@ use crate::error::{FederationError, SingleFederationError};
 use crate::query_plan::operation::normalized_field_selection::NormalizedField;
 use crate::query_plan::operation::normalized_inline_fragment_selection::NormalizedInlineFragment;
 use crate::query_plan::operation::NormalizedSelectionSet;
+use crate::schema::field_set::parse_field_set;
 use crate::schema::position::{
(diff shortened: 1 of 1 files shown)


src/query_graph/mod.rs
```diff
@@ -695,11 +696,56 @@ impl QueryGraph {
...
+                return Err(FederationError::internal(format!(
+                    "get_locally_satisfiable_key must be called on a composite type, got {}",
+                    node.type_
+                )));
+            }
+        };
+        let schema = self.schema_by_source(&node.source)?;
+        let Some(metadata) = schema.subgraph_metadata() else {
+            return Err(FederationError::internal(format!(
+                "Missing metadata on subgraph {}",
```
lrlna, line 721: Have we been keeping the error messages exactly the same in the port, or have we been changing them to be more aligned with Rust error messages? This one makes sense to me though.

lrlna (approved):

src/query_graph/mod.rs
goto-bus-stop: typically the same. the intentional difference here is to write "subgraph metadata" instead of "federation metadata" to align with the new name of the rust struct. i updated it to invent less new stuff in b139d136e24ac305110814779511b1960ad96bcd :)

PR #273, opened 2024-04-24
Port has_an_implementation_with_provides()

lrlna:
This is used in `OpGraphPath::advance_with_operation_element()`.

Original in JS: https://github.com/apollographql/federation/blob/bbf83946800d0e6f83307e1941cd7052f8a3c38d/query-graphs-js/src/graphPath.ts#L2213-L2224

### DIFF
--- a/src/query_graph/mod.rs
+++ b/src/query_graph/mod.rs
@@ -13,7 +13,6 @@ use indexmap::{IndexMap, IndexSet};
 use petgraph::graph::{DiGraph, EdgeIndex, EdgeReference, NodeIndex};
 use petgraph::visit::EdgeRef;
 use petgraph::Direction;
-use std::any::type_name;
 use std::fmt::{Display, Formatter};
 use std::hash::Hash;
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


src/query_graph/mod.rs
```diff
@@ -723,9 +723,44 @@ impl QueryGraph {
...
+            )));
+        };
+
+        let provides_directive_definition = metadata
+            .federation_spec_definition()
+            .provides_directive_definition(schema)?;
+        let type_name = interface_field_definition_position.parent();
+        let interface_type = type_name.get(schema.schema())?;
+
+        if let Some(implements_interface) = interface_type.implements_interfaces.iter().next() {
```
goto-bus-stop, line 743: This looks the wrong way around. In JS we are checking all object types that implement this interface, but `.implements_interfaces` contains the super-interfaces of this interface.

The Rust codebase has a `schema.possible_runtime_types(position)?` method for accessing that information.

It should also be a loop so we check every implementation and not just the first

goto-bus-stop (changes requested):

src/query_graph/mod.rs
lrlna: ohhhh i didn't find this function! with that this looks a lot better now

goto-bus-stop (approved):
Perfect thank you!

PR #274, opened 2024-04-24
Add a temporary error code definition for internal errors

lrlna:
We are using internal errors *a ton* inside the query planner, and while `ErrorCode::definition` is not yet used, I don't want to accidentally have us run into a panic with error codes. There is a follow up ticket for documenting this when we deal with composition later on - https://apollographql.atlassian.net/browse/FED-190

### DIFF
--- a/src/error/mod.rs
+++ b/src/error/mod.rs
@@ -1127,6 +1127,12 @@ lazy_static! {
             replaces: &[],
         }),
     );
+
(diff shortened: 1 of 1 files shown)


goto-bus-stop (approved):
Seems good, I'm not sure if this is supposed to be used for anything other than docs generation, we could probably make this much much much more ergonomic and better with some simple macros in the near future.

lrlna:
I am pretty sure it's not in the hot path, but there are like 20 internal errors in the query planner 🙈  so just in case

PR #277, opened 2024-04-26
FED-47: Finish porting `compute_best_plan_from_closed_branches`

duckki:
~Mostly finished, but still in progress. I wanted to get some feedback in a few areas.~

This PR is now ready for review.

####  `OpPathTree::merge(&self, ...) -> Self`

~I needed `OpPathTree::merge(&self, ...) -> Self`, but `merge` name is taken.
We already have `PathTree::merge(self: &Arc<Self>, ...) -> Arc<Self>`.
Probably, we can refactor the common part out. (@SimonSapin What do you think? Would there be a better way?)
But, how would you like to name the new method? I tentatively chose `merge_raw`.~

Update: I decided to use the existing `PathTree::merge` method and it worked.

#### `impl Clone for FetchDependencyGraph`

~I added a simple cloning, but it's probably wrong, since the JS version looks a lot more complicated.~

Update: Sachin confirmed the derived Clone implementation would suffice.

### Debug logging

~JS code has `debug.log(...)` lines. How do we want to port those?~

Update: I left the logging code, but commented it out. We can decide what to do after rover integration.


src/query_graph/graph_path.rs
```diff
@@ -430,6 +430,25 @@ impl Display for OpGraphPathContext {
 #[derive(Clone)]
 pub(crate) struct SimultaneousPaths(pub(crate) Vec<Arc<OpGraphPath>>);
 
+impl SimultaneousPaths {
+    pub(crate) fn fmt_indented(&self, f: &mut Formatter<'_>, indent: &str) -> std::fmt::Result {
```
SimonSapin, line 434: Is this only for potential debug logging? If not and it is in a hot path: it’s possible to refactor this to not allocate intermediate strings by writing components to `&mut Formatter` one by one. `src/query_plan/display.rs` is an example of this pattern. (`current_indent_level` there enables arbitrarily-nested indentation, but is not needed if there’s exactly one ident level.)

duckki: It's used for debug logging. However, logging at runtime is still a hot path. Let me improve it a bit.

BTW, the indentation is not used for now, but it was used somewhere else in the original TypeScript implementation. So, I tried to port the indentation feature just in case.

src/query_graph/graph_path.rs
```diff
@@ -438,6 +455,12 @@ impl std::fmt::Debug for SimultaneousPaths {
     }
 }
 
+impl std::fmt::Display for SimultaneousPaths {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        self.fmt_indented(&mut IndentedFormatter::new(f))
+    }
+}
+
```
lrlna, line 463: I'd just have the fmt_indented straight up in here, I don't think it needs an extra Impl block.

I also feel a bit strange about matching on sliced elements. A nicer pattern would be to do a `.split_first()`.

```suggestion
impl std::fmt::Display for SimultaneousPaths {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        let mut state = IndentedFormatter::new(f);

        if let Some((first, elements)) = self.0.split_first() {
            state.write_fmt(format_args!("{{ {first} }}"))?;
            state.write("{")?;
            write_indented_lines(&mut state, elements, |f, elem| f.write(elem))?;
            state.write("}")
        } else {
            state.write("<no path>")
        }
    }
}
```

src/query_graph/graph_path.rs
```diff
@@ -430,6 +431,22 @@ impl Display for OpGraphPathContext {
...
+            _ => {
+                f.write("{")?;
+                write_indented_lines(f, &self.0, |f, elem| f.write(elem))?;
+                f.write("}")
+            }
+        }
+    }
+}
+
```
lrlna, line 449: I'd just have this as part of Display impl below (see the following suggestion)
```suggestion
```

src/query_plan/display.rs
```diff
@@ -7,29 +7,40 @@ pub(crate) struct State<'fmt, 'fmt2> {
     output: &'fmt mut fmt::Formatter<'fmt2>,
 }
 
-impl State<'_, '_> {
-    fn write<T: fmt::Display>(&mut self, value: T) -> fmt::Result {
+impl<'a, 'b> State<'a, 'b> {
+    pub(crate) fn new(output: &'a mut fmt::Formatter<'b>) -> State<'a, 'b> {
```
lrlna, line 11: This is more for a hygiene follow up in the future. It seems like we now how specific display impls in `query_plan` and `query_graph` modules. I like that `query_plan` groups it all together in `display.rs`, `query_graph` doesn't have that. I think we can have this `State` live outside of the two modules (straight up in `src`?) and have `display.rs` be implemented in both.

src/query_plan/generate.rs
```diff
@@ -11,6 +11,24 @@ struct Partial<Plan, Element> {
     index: Option<usize>,
 }
 
+// PORT_NOTE: In TypeScript code, `generate_all_plans_and_find_best` takes three closures
+// as arguments. However, more than one capture the QueryPlanningTraversal object and
+// one of them mutably captures it, which is not allowed in Rust. Therefore, we use a trait
+// that implements all three methods.
+pub trait GenerateContext<Plan, Element> {
```
lrlna, line 18: Maybe something like `PlanContext`? `GenerateContext` seems a bit vague here.

src/query_plan/query_planning_traversal.rs
```diff
@@ -446,7 +470,154 @@ impl<'a> QueryPlanningTraversal<'a> {
...
+                    .outer_self
+                    .updated_dependency_graph(&mut updated_graph, &tree);
+                if result.is_ok() {
+                    let updated_tree = plan_tree.merge(&tree);
+                    (updated_graph, updated_tree)
+                } else {
+                    // Failed to update. Return the original plan.
+                    (updated_graph, plan_tree.clone())
+                }
+            }
```
lrlna, line 573: Having read through this impl of the trait's `.add`, I really don't think it needs to be a trait. This add can just live as a function on this struct instead.

lrlna (approved):
I am going to "approve" so you're not stuck without a review due to timezones. But it would be good to change the `GenerateContext` trait into methods directly on the `QueryPlanningTraversal` impl. It'll simplify some of the definitions and the usage in `generate_all_plans_and_find_best`.

duckki:
@lrlna Thanks for the review. I think I addressed most of your comments. But, I kept `fmt_indented` and the trait for `generate_all_plans_and_find_best` argument. I hope that's ok with you. I'll go ahead and merge.

src/query_graph/graph_path.rs
dariuszkuc: I agree with Iryna that we should just move that logic to the display trait as even if `fmt_indented` was used elsewhere in JS codebase -> if display is just referencing the `fmt_indented` in all other places anyway we could now just rely on the display to do the same thing.

lrlna: It's an anti-pattern in Rust to have formatting done in a non-Display or non-Debug implementations. The Rust implementation should be using those whenever a formatted version of a struct is needed, and not a one-off method. It's an expectation of the ecosystem.

PR #282, opened 2024-04-26
create graph nodes for connect sources

nicholascioli:
This PR adds the logic necessary for creating all needed graph nodes for the query plan engine to later use during execution.

### DIFF
--- a/src/sources/connect/query_graph.rs
+++ b/src/sources/connect/query_graph.rs
@@ -572,8 +572,8 @@ mod tests {
                 SourceFederatedAbstractFieldQueryGraphEdge,
                 SourceFederatedConcreteFieldQueryGraphEdge, SourceFederatedConcreteQueryGraphNode,
                 SourceFederatedEnumQueryGraphNode, SourceFederatedQueryGraph,
-                SourceFederatedScalarQueryGraphNode, SourceFederatedTypeConditionQueryGraphEdge,
-                SourceId,
+                SourceFederatedScalarQueryGraphNode, SourceFederatedSourceEnteringQueryGraphEdge,
+                SourceFederatedTypeConditionQueryGraphEdge, SourceId,
             },
         };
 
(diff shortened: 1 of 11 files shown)


src/sources/connect/query_graph.rs
```diff
@@ -0,0 +1,812 @@
...
+                    .iter()
+                    .map(move |d| (object_field, field, d))
+            })
+            .flatten_ok()
+            .map_ok(|(object_field, field, directive)| {
+                ConnectDirectiveArguments::try_from(directive)
+                    .map(move |args| (object_field, field, args))
+            })
+            .flatten_ok()
+            .try_collect()?;
```
lennyburdette, line 68: let's move this into a separate function — i think i'd like to process the connect and source directives into a normalized representation, _and then_ validate and process into a query graph separately

src/sources/connect/query_graph.rs
```diff
@@ -0,0 +1,812 @@
...
+                    ConnectFederatedConcreteQueryGraphNode::ConnectParent {
+                        subgraph_type: object.parent().clone(),
+                    },
+                ),
+            )?;
+
+            // Process the field, constructing the rest of the graph for its selections
+            // TODO: What should we do if the selection is empty? Make a selection with all fields?
+            let field_node =
+                process_selection(args.selection.unwrap(), field, &subgraph.schema, builder)?;
```
lennyburdette, line 84: if we're going to traverse the selection AST as-in, then we can delete the `From<Selection> for GraphQLSelection` code here https://github.com/apollographql/federation-next/blob/connectors/src/sources/connect/selection_parser.rs#L734

i think this makes sense because we'll want to keep track of the complete JSONSelection component with each edge (not just the "left side" of the selection that relates to the schema)

src/sources/connect/query_graph.rs
```diff
@@ -0,0 +1,789 @@
...
+        subgraph: ValidFederationSubgraph,
+        builder: &mut impl IntraSourceQueryGraphBuilderApi,
+    ) -> Result<(), FederationError> {
+        // Extract the connects from the schema definition and map them to their `Connect` equivalent
+        // Note: We need to keep the object_field as well as the arguments since it contains important
+        // type information of the original schema.
+        // TODO: Validate that we don't have connect directives on anything but object fields.
+        let connectors = Connector::from_valid_schema(&subgraph.schema, subgraph.name.into())?;
+
+        for (id, connect) in connectors {
```
lennyburdette, line 39: `builder.add_and_set_current_source(SourceId::Connect(id.clone()))?;`

> 4a. Compute the source ID. Note in the directive position that the directive name is the name in schema, and the index is its place among other directives of the same name in the field’s directive list (so the index is 0 unless there are multiple @connect directive applications).

src/sources/connect/query_graph.rs
nicholascioli: Will be done by validation work separate to this PR

PR #284, opened 2024-04-26
FED-189: port extractDeferFromOperation

duckki:
Implemented `extract_defer_from_operation` function.

Even though most of its functionality won't be used until `@defer` is actually implemented, it's called from `compute_nodes_for_op_path_element` and we would've needed some implementation anyways. So, I ported all of it.

### DIFF
--- a/src/query_plan/fetch_dependency_graph.rs
+++ b/src/query_plan/fetch_dependency_graph.rs
@@ -192,25 +192,18 @@ pub(crate) struct DeferTracking {
 #[derive(Debug, Clone)]
 pub(crate) struct DeferredInfo {
     pub(crate) label: DeferRef,
-    pub(crate) path: FetchDependencyGraphPath,
+    pub(crate) path: FetchDependencyGraphNodePath,
     pub(crate) sub_selection: NormalizedSelectionSet,
     pub(crate) deferred: IndexSet<DeferRef>,
     pub(crate) dependencies: IndexSet<DeferRef>,
 }
 
 // TODO: Write docstrings
-#[derive(Debug, Clone)]
-pub(crate) struct FetchDependencyGraphPath {
-    pub(crate) full_path: OpPath,
-    pub(crate) path_in_node: OpPath,
-    pub(crate) response_path: Vec<FetchDataPathElement>,
-}
-
 #[derive(Debug, Clone, Default)]
 pub(crate) struct FetchDependencyGraphNodePath {
-    full_path: Arc<OpPath>,
-    path_in_node: Arc<OpPath>,
-    response_path: Vec<FetchDataPathElement>,
+    pub(crate) full_path: Arc<OpPath>,
+    pub(crate) path_in_node: Arc<OpPath>,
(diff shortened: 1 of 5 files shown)


duckki:
About the `FetchDependencyGraphPath` => `FetchDependencyGraphNodePath` change, see [my Slack discussion with Sachin](https://apollograph.slack.com/archives/C05263RUETS/p1714178621288419).

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -192,25 +192,18 @@ pub(crate) struct DeferTracking {
...
-}
-
 #[derive(Debug, Clone, Default)]
 pub(crate) struct FetchDependencyGraphNodePath {
-    full_path: Arc<OpPath>,
-    path_in_node: Arc<OpPath>,
-    response_path: Vec<FetchDataPathElement>,
+    pub(crate) full_path: Arc<OpPath>,
+    pub(crate) path_in_node: Arc<OpPath>,
+    pub(crate) response_path: Vec<FetchDataPathElement>,
```
TylerBloom, line 206: Let's limit visibility for fields that don't need to be made public.
```suggestion
    path_in_node: Arc<OpPath>,
    response_path: Vec<FetchDataPathElement>,
```

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -2429,13 +2422,56 @@ fn compute_input_rewrites_on_key_fetch(
...
+            path_to_defer_parent: updated_path_to_defer_parent.into(),
+            // Following fields are identical to those of `defer_context`.
+            current_defer_ref: defer_context.current_defer_ref.clone(),
+            active_defer_ref: defer_context.active_defer_ref.clone(),
+            is_part_of_query: defer_context.is_part_of_query,
+        };
+        return Ok((Some(operation.clone()), updated_context));
+    };
+
+    let Some(ref updated_defer_ref) = defer_args.label else {
```
TylerBloom, line 2448: Use the existing `label` method:
```suggestion
    let Some(updated_defer_ref) = defer_args.label() else {
```

src/query_plan/fetch_dependency_graph.rs
```diff
@@ -2429,13 +2422,56 @@ fn compute_input_rewrites_on_key_fetch(
...
+        // PORT_NOTE: The original TypeScript code has an assertion here.
+        return Err(FederationError::internal(
+            "All defers should have a label at this point",
+        ));
```
TylerBloom, line 2452: Also, this is more stylistic, but using `Option::ok_or_else` + a try operator (i.e. `?`) is often an easier-to-read way of short-circuiting a function that returns a `Result` when you have an `Option`. Not a necessary change, as both ways will compile to basically the same thing, but I figured I would mention it.

TylerBloom (changes requested):
Looks good overall, just a few places to be cleaned up, I think. :)

duckki:
@TylerBloom Thanks for the review! I just accepted your suggestions and pushed it.

